PR #22, opened 2019-09-17
Add appveyor configuration

Ybalrid:
This adds an appveyor.yml file that configure a CI build for MSVC (x64 2013, we can activate more versions by un-commenting the matrix blocks in the yml file) on the AppVeyor service.

AppVeyor needs to be linked from somebody's account with access to OGRECave organization 

After merging this, just create an account on AppVeyor (https://www.appveyor.com/ - you can use GitHub OAuth), link your GitHub to it, and add the OGRECave/ogre-next repo to AppVeyor.

It will build all commits and pull requests on this repo. It's slow but it works...

OpenSouce projects have free unlimited builds on AppVeyor :wink:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,26 @@
+version: 2.2.{build}
+
+environment:
+  matrix:
+#    - PLATFORM: x86
+#      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+#      BUILD_SCRIPT: build_ogre_Visual_Studio_15_2017_Win32.bat
+#    - PLATFORM: x86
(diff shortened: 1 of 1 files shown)


darksylinc:
That looks so easy!

I should update the build scripts to pull the repo from git rather than hg

darksylinc:
Do you know if AppVeyor supports incremental builds? i.e. run the script again over an existing build file? That would significantly decrease build times

Ybalrid:
@darksylinc There are ways to cache some files, but I'm not sure we can make it do incremental builds. One thing we can do is cache the dependencies at least, not building OIS and FreeImage and SDL2 and Free-type each time would shave off quite a lot of time from the CI execution

PR #38, opened 2019-11-03
Docs: point out render order if DisableSort is used

mimon:
Based on what @darksylinc  wrote in https://forums.ogre3d.org/viewtopic.php?f=2&t=94628#p543341

### DIFF
--- a/OgreMain/include/OgreRenderQueue.h
+++ b/OgreMain/include/OgreRenderQueue.h
@@ -209,6 +209,11 @@ namespace Ogre {
                               const MovableObject *pMovableObject );
 
         /** Add a renderable (Ogre v2.0, i.e. Items; they use VAOs) object to the queue.
+        @remarks
+            If sorting mode is set to DisableSort (RenderQueue::setSortRenderQueue)
+            for the given renderQueueId, then the order in which renderables are
+            added via RenderQueue::addRenderableV2 (and V1) determines the
+            render order (the first added get rendered first) within that renderQueueId.
         @param threadIdx
             The unique index of the thread from which this function is called from.
             Valid range is [0; SceneManager::mNumWorkerThreads)


darksylinc:
I find that description to be very easy to misunderstand, as it could be interpreted that DisableSort uses the order in which Items (or Entities) were created, which is not true (but it will often coincide but not always, adding confusion).

DisableSort honours the order in which addRenderableV2 (or addRenderableV1) was called. That's it.

If you're manually manipulating the RenderQueue then yes, this is useful information.

But if you're relying on the standard Ogre rendering paths (i.e. drawing regular Items) then some of these objects are added from a multithreaded context.

Order in which Items were created will often coincide because each thread is currently splitting all MovableObjects into chunks, and the threads later get collected in order.

But if you start destroying Items and then add more; released slots are recycled for new objects and that's when the order breaks. But this is an implementation side effect that is not enforced.

The only order with DisableSort that can be safely guaranteed is the order in which addRenderableV1/addRenderableV2 calls are made.

mimon:
> I find that description to be very easy to misunderstand, as it could be interpreted that DisableSort uses the order in which Items (or Entities) were created

I understand your concern. However, since this piece of docs is specifically for `RenderQueue::addRenderableV2()` and exactly describes the behaviour of the `RenderQueue` when `DisableSort` is set, why would I confuse it with the behaviour of e.g. `SceneManager::createItem()`? I can't see the potential for misunderstanding 🤔

darksylinc:
Just slightly reword the paragraph to say '..then the order in which renderables are added via RenderQueue::addRenderableV2 (and V1) determines...'

mimon:
> Just slightly reword the paragraph to say '..then the order in which renderables are added via RenderQueue::addRenderableV2 (and V1) determines...'

👍 I've updated and pushed the new changes. @darksylinc

PR #42, opened 2019-11-08
Fixed compiler issues for gcc 9.2 and FreeBSD-13.0-CURRENT

Stazer:
Hi,

first of all thank you for all your great work! I really appreciate it.

I know there probably no use case for it since only very few people use FreeBSD as their day-to-day operating system and probably less people are using it for developing applications with Ogre. 
Nevertheless I needed that possibility and thought it might be helpful to share my changes I made to get Ogre running with it. Although I was able to compile and successfully run all samples I would not suggest it to mark FreeBSD as supported platform. There are only a few changes I made, so there is definitely no impact on any other supported platform. 

Best Regards


OgreMain/src/OgrePlatformInformation.cpp
```diff
@@ -639,7 +639,7 @@ namespace Ogre {
     {
         uint32 numLogicalCores = 0;
 
-#if OGRE_PLATFORM == OGRE_PLATFORM_LINUX || OGRE_PLATFORM == OGRE_PLATFORM_ANDROID
+#if OGRE_PLATFORM == OGRE_PLATFORM_LINUX || OGRE_PLATFORM == OGRE_PLATFORM_ANDROID || OGRE_PLATFORM_FREEBSD
```
darksylinc: This line here is causing trouble.

`OGRE_PLATFORM_FREEBSD` should be `OGRE_PLATFORM == OGRE_PLATFORM_FREEBSD`

darksylinc:
Hi!

The commit is almost alright, there is a line I pointed out in Github that is causing trouble, in OgrePlatformInformation.cpp around line 643

OgreMain/src/OgrePlatformInformation.cpp
Stazer: Fixed it in 4d3aa2a.

darksylinc:
Thanks for your contribution!
I was impressed "it just works" on FreeBSD. It's not an OS I'm used to.

PR #46, opened 2019-11-23
Export MetalPlugin and GL3PlusPlugin classes

mimon:
On clang, macOS mojave, unless MetalPlugin is exported, dynamic linking yields

```
Undefined symbols for architecture x86_64:
  "Ogre::MetalPlugin::MetalPlugin()", referenced from:
      flexbox::application::application() in application.cpp.2.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Note: The plugin is not loaded at run-time in this case.

Perhaps I should set the other render system plugins to be exported as well?

### DIFF
--- a/RenderSystems/GL3Plus/include/OgreGL3PlusPlugin.h
+++ b/RenderSystems/GL3Plus/include/OgreGL3PlusPlugin.h
@@ -35,7 +35,7 @@ namespace Ogre
 {
 
     /** Plugin instance for GL3Plus Manager */
-    class GL3PlusPlugin : public Plugin
+    class _OgreGL3PlusExport GL3PlusPlugin : public Plugin
     {
     public:
         GL3PlusPlugin();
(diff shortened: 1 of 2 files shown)


darksylinc:
Hi!

It's weird that I didn't catch this bug (works fine for me), but you're right.
Btw could you replace the '_OgreExport' for '_OgreMetalExport' (which is the right macro to use within the MetalRenderSystem)?

Cheers

mimon:
> Hi!
> 
> It's weird that I didn't catch this bug (works fine for me), but you're right.
> Btw could you replace the '_OgreExport' for '_OgreMetalExport' (which is the right macro to use within the MetalRenderSystem)?
> 
> Cheers

Sure!

What about the GL3PlusPlugin, it's also missing the export directive, should it be left as-is?

mimon:
> Hi!
> 
> It's weird that I didn't catch this bug (works fine for me), but you're right.
> Btw could you replace the '_OgreExport' for '_OgreMetalExport' (which is the right macro to use within the MetalRenderSystem)?
> 
> Cheers

I have updated the branch with new changes @darksylinc

darksylinc:
Yeah, add if to GL3+ too.

Thanks for looking into it!

mimon:
> Yeah, add if to GL3+ too.
> 
> Thanks for looking into it!

Code updated! @darksylinc

PR #50, opened 2019-11-24
Enable clickthrough in Metal OSX

Unix4ever:
Problem:
Click on an unfocused window does not work. For example if you want to
press the button it won't handle mouse down.
With SDL2 you can fix that by setting
`SDL_HINT_MOUSE_FOCUS_CLICKTHROUGH` and it will properly set up
clickthrough for `NSWindow`.

But OGRE adds yet another NSView on top of it which does not handle
clickthrough.

Solution:
Make NSView that contains Metal layer mimic the behaviour of parent
NSView.
This way setting `SDL_HINT_MOUSE_FOCUS_CLICKTHROUGH` starts working as
it should.

### DIFF
--- a/RenderSystems/Metal/src/Windowing/OSX/OgreMetalView.mm
+++ b/RenderSystems/Metal/src/Windowing/OSX/OgreMetalView.mm
@@ -82,6 +82,15 @@ THE SOFTWARE.
     return self;
 }
 
+- (BOOL)acceptsFirstMouse:(NSEvent *)theEvent
+{
+    if(self.superview == nil) {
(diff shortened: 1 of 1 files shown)


Unix4ever:
If you merge this PR, can you also add this change to 2-1 branch? It's the version I use.

darksylinc:
Added to 2.1 as requested: https://github.com/OGRECave/ogre-next/commit/2bac7a05895b6bc0145dd8e8b641de5fd1c4c432

Unix4ever:
Thanks! :+1:

PR #51, opened 2019-11-28
Do XSync right after calling XDestroyWindow

Unix4ever:
I've hit the following corner case:
- create SDL2 window.
- create Ogre render window and specify `parentWindowHandle`.
- close and destroy Ogre render window.
- close SDL2 parent window.

Get `BadWindow` error due to async XLib nature.
Adding `XSync` makes window destructor more predictable.

Again: would be great to get this PR merged into 2.1 as well.


darksylinc:
So THAT'S what was missing?

It drove me nuts for a few years!

PR #58, opened 2019-12-17
WIP: Updating docs to match build scripts

DavidYKay:
So far, I've changed `#DownloadingOgreCommon` and `#BuildingOgreLinux` to match the commands from the build script.

GUI directions still need to be updated.

1. Can someone critique my approach and let me know if I need to change anything?
2. The GUI directions are woefully out of date, at least on Linux. Should we nix them? Update them? What's the right path forward?

Thanks,
David

### DIFF
--- a/Docs/src/SettingUpOgre/SettingUpOgreLinux.md
+++ b/Docs/src/SettingUpOgre/SettingUpOgreLinux.md
@@ -22,11 +22,10 @@
 # Building Dependencies {#BuildingDependenciesLinux}
 
 ```sh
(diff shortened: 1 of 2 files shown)


darksylinc:
1. Looks good. Thanks!
2. Good point. I didn't realize we're still pointing to Mercurial. All platforms are in the same state. On Linux & Mac it's clear to use git CLI commands because that's the best experience there. On Windows I've got no idea what people use for Git, the UI landscape is really fragmented and there is no really good tool

PR #65, opened 2020-01-10
Fix crash when a dynamic probe is destroyed

TaaTT4:
The crash happens during update, after having removed a dynamic probe, because the content of `mDirtyProbes ` _doesn't match_ the content of `mProbes`. And then this cause that a no more existent probe is tried to render.


darksylinc:
Merged manually due to a conflict
Thanks for the fix!!!

PR #67, opened 2020-01-16
Pad cubemap probe area and shape in the same way

TaaTT4:
Not doing so, can cause "lit noise" with per-pixel PCC in some scenario. E.g., when the same `Aabb` is used for `area` as well as `probeShabe` in `CubemapProbe::set` method and a pixel to be lit resides on probe shape edges.
picturemessage_ieqxfddj p2p


darksylinc:
With picture of the error and everything! I love it!

PR #77, opened 2020-03-05
HiddenAreaMeshVR: make config devicename matching caseinsensitive

peetCreative:
My Vive gives me device name VIVE_PRO MV.
So I made the matching uppercase.


darksylinc:

Issue #53, opened 2019-12-03
OpenVR Sample: Latest won't run, older versions have rendering glitch


DavidYKay:
#### System Information
- Ogre Version: 2.2 - 8579b3508300ab3450c556c3aae0a5a22f95468e
- Operating System / Platform: Gentoo Linux 64-bit
- RenderSystem: GL3+
- GPU: NVIDIA GeForce GTX 1080

#### Detailed description

Hi guys,

Thanks very much for your work on OGRE. 

I've used it on a few toy applications in the past and am currently integrating it into a VR application.

I'm having trouble getting the Tutorial_OpenVR sample to work correctly.

##### Case 1: Latest Code - Crash

The latest code, 52a6f47c5f21a4a9c113b4e9d2167502bfdbabff, crashes with the following error:

`An exception has occured: OGRE EXCEPTION(5:ItemIdentityException): Compute Job with name [Hash 0x704694c7] not found in HlmsCompute::findComputeJob at [OGRE_ROOT]/OgreMain/src/OgreHlmsCompute.cpp (line 557)`

After defining `OGRE_IDSTRING_ALWAYS_READABLE 1`, the IdString in question turns out to be: `VR/RadialDensityMaskReconstruct`.

I've attempted to pass in a computer shader, using `HlmsComputeJob::createComputeJob` and guessing that `/Samples/Media/Compute/VR/Foveated/RadialDensityMaskReconstruct_cs.glsl` is the file in question, but I haven't gotten this to work yet.

##### Case 2: Old Code - Screen Door

In order to find a working example, I dug around in the git history and got these two commits to run, albeit with a significant rendering glitch.

ea3b3f17260fa2e844ef29eb3cc26cd72e2f8327, 075d36b739ba9c66ff730de28c259e85b8c7b198 -> Both run, but with many visual artifacts.

![pixelated](https://user-images.githubusercontent.com/171287/70031294-c1757c00-155f-11ea-9a8b-ebdc83b1a50b.png)

##### Case 3: Hybrid Code - Crash

Finally, I tried combining the latest OGRE code with the "working" Tutorial_OpenVR code,

OGRE commit: 52a6f47c5f21a4a9c113b4e9d2167502bfdbabff 
OpenVR_Tutorial commit: ea3b3f17260fa2e844ef29eb3cc26cd72e2f8327

But this produces the same results as first case, crashing with `ItemIdentityException` in `HlmsCompute::findComputeJob`.

#### Ogre.log & Callstack

[...]

PR #81, opened 2020-04-02
Openvr sample

peetCreative:
It does not solve #53, but now one can disable the RDM with a compile flag OGRE_BUILD_SAMPLES_OPEN_VR_RDM
And for anybody, who is interested in solving the problem, but lacks a HMD. The sample compiles without OpenVR. The RDM problem still persists.


peetCreative:
Last change actually solves #53.
However I guess MSAA is not defined in TextureGpu.
Maybe you can comment on what was the problem here.

darksylinc:
Hi!

Your PR showed me what was needed to be done (and finally gave me the courage to deal with this issue) so it was definitely not in vain!

Using a macro to toggle RDM was overkill because macros are useful when:

- A feature is too expensive to be included by default (i.e. don't penalize everyone for a minor feature someone needs). This typically happens because an optional feature needs to be used in a "hot" path (e.g. the main render loop which may iterate over dozen of thousands of items); or significant extra memory overhead is required
- The code diverges significantly

In the case of RDM, this was not the case as it can even be turned on/off at runtime without any major hit and minimal code changes.

I also fixed a race condition (but I doubt it will fix your issue... though maybe you're lucky?) and implemented a NullCompositorListener so that I can test without the headset more easily and the render paths are now much more similar.

Thanks for your contribution!

PR #86, opened 2020-04-16
Concentric shadow setup are now stable for position changes. added nu…

yoyTeam:
…mStableSplits to createShadowNodeWithSettings


darksylinc:
Thanks!

I've made a small math optimization since you already had the position in lights space in which you could work with:
https://github.com/OGRECave/ogre-next/commit/f948c8761cd70732a98bd0119de6ffe9dd29f69e

I also applied your code to FocusedCameraSetup https://github.com/OGRECave/ogre-next/commit/cab94ddf673615899981c81b15cc6407b1af5838 he effect is much weaker than in ConcentricShadowCamera but at least it's something

If testing shows there are no further issues we'll just merge the branch back to v2-2-irradiance-field

PR #89, opened 2020-04-25
Fix doxygen publish

mimon:
This should do it 😄  🤞

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,7 +36,7 @@ jobs:
           doxyfile-path: './Doxyfile'
 
       - name: Publish # Only on master branch
-        if: github.ref == 'master'
+        if: github.ref == 'refs/heads/master'
         uses: peaceiris/actions-gh-pages@v3
         with:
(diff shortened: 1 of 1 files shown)


mimon:
Hm, I see there's a 404 now on GH pages. Not what I expected I suppose 😬 . It's weirded I had not these issues in my own test repo.

mimon:
Looks like it's not serving from the `html/` directory. So one needs to go to https://ogrecave.github.io/ogre-next/html/ to see the page

mimon:
Perhaps [this line](https://github.com/OGRECave/ogre-next/blob/master/CMake/Templates/html.cfg.in#L898) should be changed to `./` instead

darksylinc:
I force-push reverted that change since we can't have broken links on documentation; at least until we find the proper solution

mimon:
> I force-push reverted that change since we can't have broken links on documentation; at least until we find the proper solution

OK. I'll have a look tomorrow, but I'm pretty sure this is the fix:
https://github.com/OGRECave/ogre-next/pull/89#issuecomment-619414327

Issue #92, opened 2020-04-29
Use relative file postioning in includes


Caellian:
I'm completely new to OGRE and I've been setting up my project according to the official tutorial [here](https://ogrecave.github.io/ogre/api/latest/setup.html). I'm using Arch Linux. I know ogre-next is not very friendly for new users but I'd prefer using ogre-next for learning as it'll force me to get more familiar with how it works. I used setup scripts to setup ogre and installed it using generated ninja files.

`#include <OGRE/Ogre.h>` gives me an error [here](https://github.com/OGRECave/ogre-next/blob/6a814a02e69dfe3c006baf8ec6399aeb611452ca/OgreMain/include/Vao/OgreVertexBufferPacked.h#L32) saying that `"Vao/OgreBufferPacked.h"` file does not exist.
I checked locally, it's does (`/usr/local/include/OGRE/Vao/OgreBufferPacked.h`).

From looking at sources I infer that `OGRE` is actually the root of included files - by which I mean \#include <Ogre.h> would actually include that file in project sources. Problem is that on Linux, it is placed in `/usr/local/include/OGRE` file which means that all ogre includes have to be prefixed with `OGRE/`. Prepending `OGRE/` to header include paths fixes this issue. Moving the includes outside of 'OGRE' subdirectory would too, but I don't want to pollute my include directory with files unmanaged by the package manager.
What I also find confusing is the usage of `"` for absolute include paths. From my understanding of C++, `"` is conventionally used for relative include paths and angled brackets for paths relative to searched include directories (although `"` falls back to that behaviour so I might be just nitpicking now). Using relative include paths would be much better in cases like these and would make include files resilient to inconsistent placement.

Placing ogre headers directly inside include directory is bad.
Here are some possible solutions:
- Consider using relative paths in `#include "something.h"` statements, this would make setting up SDK much simpler.
- Consider moving includes into OGRE subdirectory (requires much larger refactoring, breaks stuff, not really an option)

While this might be easy to fix with a script prepending OGRE to includes, consider the fact that lots of new users using Linux will face a similar problem and patching files locally is not the best solution.
I might also be doing something wrong in which case most of the issue above doesn't make much sense.

AllianceDarkSylinc:
Hi!

Beware the tutorial you're looking at is for ogre 1.x. This branch is [ogre-next](https://ogrecave.github.io/ogre-next/api/2.1/_using_ogre_in_your_app.html).
Although there are many similarities, there will also be differencies.

[...]

PR #93, opened 2020-04-29
Fix OgreWorkarounds.h encoding

Caellian:
## Changes
- Fixed OgreWorkarounds.h encoding
- Replaced double quoted includes with angle brackets for system headers for clarity.

I mentioned these in #92.


darksylinc:
Thank you for the contribution!

PR #95, opened 2020-05-01
Put API docs under /api/2.2/ when publishing to gh-pages

mimon:
This will publish the html at the same path where the current api docs resides in (i.e. https://ogrecave.github.io/ogre-next/api/2.2/).

I did a build on my own repo's master and made sure it works: https://mimon.github.io/ogre-next/api/2.2/

## Caveat 1
All Ogre files in the the repo will be copied to the published `gh-pages` so the published directory will actually be something like:
```
/
  CmakeFiles/
  Components/
  ...
  api/2.2/...
  ...
```

## Caveat 2
The current main page at https://ogrecave.github.io/ogre-next will disappear. I'm not really sure how to build both doxygen and the main page.

## Conclusion
Due to the simple way gh-pages work and the simple build system I haven't been able to find a perfect clean way to keep both main page and the docs.

However, I'd recommend that you setup a custom subdomain DNS entry (e.g. `docs.ogre3d.org`) for the API docs. This will make it easier to maintain http links in the long-term and is independent of where the docs are hosted.

LMK what you think. Cheers

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,6 +14,8 @@ env:
   CMAKE_DOXYGEN_INPUT_LIST: "Components Docs/src OgreMain PlugIns RenderSystems"
   OGRE_SOURCE_DIR: "./"
   OGRE_BINARY_DIR: "./"
+  OGRE_VERSION: "2.2"
+  DOXYGEN_HTML_OUTPUT_DIR: "./2.2"
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
@@ -35,9 +37,10 @@ jobs:
         with:
           doxyfile-path: './Doxyfile'
 
-#      - name: Publish # Only on master branch
-#        if: github.ref == 'refs/heads/master'
-#        uses: peaceiris/actions-gh-pages@v3
-#        with:
-#          github_token: ${{ secrets.GITHUB_TOKEN }}
-#          publish_dir: ./api
+      - name: Publish # Only on master branch
+        if: github.ref == 'refs/heads/master'
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./
+          keep_files: true
--- a/CMake/Templates/html.cfg.in
+++ b/CMake/Templates/html.cfg.in
@@ -895,7 +895,7 @@ GENERATE_HTML          = YES
 # If a relative path is entered the value of OUTPUT_DIRECTORY will be
 # put in front of it. If left blank `html' will be used as the default path.
 
-HTML_OUTPUT            = html
+HTML_OUTPUT            = ${DOXYGEN_HTML_OUTPUT_DIR}
 
 # The HTML_FILE_EXTENSION tag can be used to specify the file extension for
 # each generated HTML page (for example: .htm,.php,.asp). If it is left blank


darksylinc:
I don't care for the main page at the time being; however do you know if the 2.1 docs get deleted too? (eg https://mimon.github.io/ogre-next/api/2.1/)

mimon:
> I don't care for the main page at the time being; however do you know if the 2.1 docs get deleted too? (eg https://mimon.github.io/ogre-next/api/2.1/)

Oh, yes :/ ... forgot about that. Hm, ok let me re-iteratate and see how I can solve that.

What do you think about setting up a new docs subdomain for ogre?

darksylinc:
> Oh, yes :/ ... forgot about that. Hm, ok let me re-iteratate and see how I can solve that.

For what I can witness the straightforward path would be to run a bash script which builds the APIs, checkouts a different branch after that, and builds the new API, then both are published.

> Btw, what do you think about setting up a new docs subdomain for ogre?

I'm not handling subdomain stuff so I don't know. Surely docs.ogre3d.org would confuse both Ogre1 and Ogre-next though, so a different name would be needed.

However I understand a subdomain would fix the hosting problem; but not the actual links (i.e. the html files) thus I fail to see how beneficial it is.

mimon:
> However I understand a subdomain would fix the hosting problem; but not the actual links (i.e. the html files) thus I fail to see how beneficial it is.

Well, the doxygen relative links will follow with the hostname so no problems there. But, yes, it has to still be a doxygen structure for it to work without an URL-rewriter.

A subdomain will at least be an improvement and not a regression. Especially now when ogre-next is released it'd be good to think about URL infrastructure to keep links alive for as long as possible. I understand it requires a bit of time and work tho, and stuff work as it is for now. I merely wanted to bring up the subject to see if it was something you had thought about.

mimon:
@darksylinc I've updated this PR to solve building v2-1 docs.

This works by setting the `keep_files: true` in the publish step in `main.yml`. Hence, docs built on `master` and `v2-1` branch can both push to `gh-pages` without overwriting each others files.

darksylinc:
Your PR is clearly working! An update to the docs was reflected in the live documentation, without breaking anything.

Kudos! Great work!

PS. I only merged it today because I found myself with available time to focus on misc stuff such as dealing with auto-gen docs (which needs me to be alert if the docs get broken)

darksylinc:
It looks like Github now insists on pushing changes made to the branches into the gh-pages branch.

If you find an easy way to avoid that, that'd make it perfect.

darksylinc:
Perhaps by running a sh script that removes all files and folders except those that start with '2.' and hidden files, so that only documentation stays pushed is the way to fix it

darksylinc:
Fixed with a Python script https://github.com/OGRECave/ogre-next/commit/04364af5a76d1e3aff87bcef892d0bb6e3f98dbe

I was misled with the keep_files thing, as per peaceiris/actions-gh-pages@v3 documentation (not jekyll), it only supports true/false

mimon:
> Your PR is clearly working! An update to the docs was reflected in the live documentation, without breaking anything.
> 
> Kudos! Great work!
> 
> PS. I only merged it today because I found myself with available time to focus on misc stuff such as dealing with auto-gen docs (which needs me to be alert if the docs get broken)

Cheers! Glad it worked!

> It looks like Github now insists on pushing changes made to the branches into the gh-pages branch.

> If you find an easy way to avoid that, that'd make it perfect.

Yes, unfortunately, all files had to be copied in order to achieve the tree structure that existing URLs expects (e.g. `http://.../api/..`). When I have time, I could try to do the following to mitigate that:

* Change Doxygen to output the compiled docs to `api/api/v2.2/` and then set `publish_dir: api/`. Not sure if `keep_files: true` is going to overwrite that though. One have to test it.

Was there any problems with copying the whole repo (apart from it being annoying ofc.)?

PR #118, opened 2020-07-25
Detect failure to create Dependency directory symlink

joealam:
If the symlink couldn't be created, show a useful error messages indicating how to solve the problem and exit.

(prevents OGRECave#117)


darksylinc:
Thanks!

Hopefully the need for this will diminish as more people update to latest Windows 10 (past Build 14972 mklink no longer requires admin privileges)

PR #119, opened 2020-07-26
Fix bug with compositor texture referenced from TextureUnitState

lingfors:
When setting the reference to the compositor texture, mFrames[0] was set to the texture name.
However, this caused the TextureUnitState to try and read the file from disk when the
TextureUnitState was loaded.

Instead, mFrames[0] should not be set to anything. mFramePtrs[0] will be correctly set
when the corresponding compositor is updated.


darksylinc:
Thanks for the fix!

It took me a little bit to approve it because I had to run the unit tests to see there were no regressions (they passed :) )

PR #120, opened 2020-07-30
Fix incorrect window texture resolution in windowed mode in OpenGl

RauchMatthias:
This fixes the issue  I described in the forum:
https://forums.ogre3d.org/viewtopic.php?f=25&t=95899


darksylinc:

PR #134, opened 2020-08-26
Fix aabb distance

lingfors:
The Array versions probably has to be fixed as well...


darksylinc:
Wow, this is old code, can't believe it went through.

Unfortunately, your code [is failing](https://ci.appveyor.com/project/MatiasNGoldberg/ogre-next/builds/34872177#L431) when built with OGRE_UNITY_BUILD.

I suspect the dreaded NOMINMAX could be the cause (Windows headers defining min and max as macros).

lingfors:
I think it's a mismatch between float and double in the call to std::max. I compile Ogre with OGRE_DOUBLE=1, and then this work.

darksylinc:
Thanks!!!

I can't believe how wrong the old version was.

This has been applied manually to fix all versions and also broadcast the fix to all branches (2.1, 2.2 and master)

PR #136, opened 2020-09-04
fix the Paging Component for v2.2

silicium42:
I have created a small patch to fix the OgrePaging Component for v2.2, because i am currently using it myself. It is currently a WIP but it seems to work ok for my use case: a custom, procedural implementation of PageStrategy, PageContent and PageProvider.
I would be thankful for any hints as to what might still be missing or wrong. The Grid2D/3DPageStrategy.cpp files are currently missing the `pass->setLightingEnabled(false);` and i don't really know what the best/correct way is to achieve this in v2.2.

### DIFF
--- a/Components/Paging/include/OgrePage.h
+++ b/Components/Paging/include/OgrePage.h
@@ -32,7 +32,7 @@ THE SOFTWARE.
 #include "OgrePagingPrerequisites.h"
 #include "Threading/OgreThreadHeaders.h"
 #include "OgreWorkQueue.h"
-
+#include "ogrestd/vector.h"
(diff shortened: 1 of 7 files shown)


darksylinc:
Do you want me to merge it right now or wait?

Given that no one is using this component, the PR can be merged immediately as it won't break anything

silicium42:
Either is fine with me, i don't think i will change it in the near future since it works for me as it is. I tested the Grid2D implementation included in the Component about a year ago with v2.1 and see no reason why it shouldn't work now with the few changes for v2.2

PR #142, opened 2020-10-03
Publish api/ directory

mimon:


darksylinc:
Please note that we've workarounded the publish issue with a python script and it seems to be working fine!

Thanks for the research though!

Eugene [fixed one last issue](https://github.com/OGRECave/ogre-next/commit/42b80e8e99cefb84518d14744d2d0eec1a5301cc) with the python script a couple days ago

PR #149, opened 2020-10-12
E2K: added initial support of MCST Elbrus 2000 CPU architecture

r-a-sattarov:
e2k (Elbrus 2000) - this is VLIW/EPIC architecture, like Intel Itanium (IA-64) architecture.

- https://en.wikipedia.org/wiki/Elbrus_2000

### DIFF
--- a/OgreMain/include/OgrePlatform.h
+++ b/OgreMain/include/OgrePlatform.h
@@ -61,9 +61,10 @@ namespace Ogre {
 #define OGRE_CPU_PPC        2
 #define OGRE_CPU_ARM        3
 #define OGRE_CPU_MIPS       4
+#define OGRE_CPU_E2K        5
 
 /* Find CPU type */
-#if defined(__i386__) || defined(__x86_64__) || defined(_M_IX86) || defined(_M_X64) || defined(_M_AMD64)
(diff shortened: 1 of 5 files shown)


EntityFX (commented):
Nice to see e2k arch here :)

darksylinc:
Hi,

We've been discussing about this PR.

Our main issue is that (so far) it doesn't add anything relevant to the project.

If the PR added a path specific only to Elbrus 2000 it could make sense. But all it does is add a new architecture macro but then proceed to treat all the e2k paths the same way as the x86 paths.

Therefore this PR only increases complexity and adds a maintenance burden, when the e2k paths could be treated under the x86 umbrella.

r-a-sattarov:
> If the PR added a path specific only to Elbrus 2000 it could make sense. But all it does is add a new architecture macro but then proceed to treat all the e2k paths the same way as the x86 paths.
> 
> Therefore this PR only increases complexity and adds a maintenance burden, when the e2k paths could be treated under the x86 umbrella.

Ok, I have simplified my PR. Could you look it, please?

Previous changes: [ogre-next-e2k-patch.txt](https://github.com/OGRECave/ogre-next/files/5489432/ogre-next-e2k-patch.txt)

PR #151, opened 2020-10-16
Fixing compile errors, when OGRE_CONFIG_DOUBLE is TRUE

niparx:
Fixed such compile errors:
- OgreScriptTranslator.cpp:8980:60: error: cannot initialize a parameter of type 'Ogre::Real *' (aka 'double *') with an rvalue of type 'float *': `if( !getReal(prop->values.front(), &passClear->mClearDepth) )`
- Components/Hlms/Pbs/src/Vct/OgreVctLighting.cpp:598:36: error: call to member function 'setManualValue' is ambiguous: `mBounceIterationDampening->setManualValue( 1.0f / Math::PI )`


darksylinc:
I'm closing it because it was [fixed by hand](https://github.com/OGRECave/ogre-next/commit/70a1889564cd46279c8212766fb3eea39868cd26).

mClearDepth is purposely a float, and there were also more build errors when using OGRE_CONFIG_DOUBLE is TRUE.

Thanks for the fixes! It would still be broken if not for this PR.

niparx:
Basically, we are using OGRE_CONFIG_DOUBLE=TRUE. I was getting only those compile errors (at least for the components, which were selected). Are there any issues, which I must be aware, related to OGRE_CONFIG_DOUBLE? Like, code is compiling, but not working as expected?

Issue #145, opened 2020-10-07
NINJA/MacOS: warning: multiple rules generate ... builds involving this target will not be correct;


crjc:
Hi,

I'm trying to set up the latest release of Ogre to build via Vcpkg on MacOS.

Ninja is complaining quite a bit so I can't get the build working. I'm not sure if anyone knows why there's multiple rules generating the following files? 

I'm hoping its a simple fix that someone could help with, I need Vcpkg builds to work on MacOS before I can get my hands dirty with ogre-next!

Log from a modified version of Vcpkg to use an old version of ninja
```
ninja: warning: multiple rules generate lib/macosx/OgreMainStatic.framework/Versions/2.2.4/Headers/OgreTagPoint.h. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
ninja: warning: multiple rules generate lib/macosx/OgreMainStatic.framework/Versions/2.2.4/Headers/OgreAabb.h. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
ninja: warning: multiple rules generate lib/macosx/OgreMainStatic.framework/Versions/2.2.4/Headers/OgreArrayAabb.h. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
ninja: warning: multiple rules generate lib/macosx/OgreMainStatic.framework/Versions/2.2.4/Headers/OgreArrayMatrix4.h. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
ninja: warning: multiple rules generate lib/macosx/OgreMainStatic.framework/Versions/2.2.4/Headers/OgreArrayMatrixAf4x3.h. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
ninja: warning: multiple rules generate lib/macosx/OgreMainStatic.framework/Versions/2.2.4/Headers/OgreArrayQuaternion.h. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
ninja: warning: multiple rules generate lib/macosx/OgreMainStatic.framework/Versions/2.2.4/Headers/OgreArraySphere.h. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
ninja: warning: multiple rules generate lib/macosx/OgreMainStatic.framework/Versions/2.2.4/Headers/OgreArrayVector3.h. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
[...]

PR #157, opened 2020-11-29
Choose between C, SSE2 and NEON at CMake Time 

crjc:
As suggested in #145, this PR merges both the `OGRE_SIMD_SSE2` and `OGRE_SIMD_NEON` flags into a single flag `OGRE_SIMD`, and CMake is now responsible for picking SSE2, NEON or C.

There were quite a few 'proxy' files, which I've deleted as they don't seem to be necessary after moving away from the macro approach.


darksylinc:
Hi!

> There were quite a few 'proxy' files, which I've deleted as they don't seem to be necessary after moving away from the macro approach.

I didn't immediately approve this PR because this is the most (and only) conflicting part.

It indeeds does seem to become unnecessary, however the folder structure was done on purpose.

Even if we purposely change to flat include files as in your PR, this change will break projects using Ogre, as they will all need to be updated to include the **right** path, and that can easily become a disaster.

Is it possible to keep the 'proxy' files?

PR #161, opened 2020-12-11
Choose SIMD implementation at cmake time

crjc:
A less drastic version of #157 (this doesn't delete proxy files or change folder structure). A motivating factor behind deleting the proxy files was to prevent multiple files having the same name so that ninja/vcpkg/Xcode would not complain, although this doesn't seem to be an issue anymore with the changes I've made.

This PR merges both the OGRE_SIMD_SSE2 and OGRE_SIMD_NEON flags into a single flag OGRE_SIMD, and CMake is now responsible for picking SSE2, NEON or C.


AllianceDarkSylinc:
Ah! I forgot that was the problem.

It IS acceptable to rename each implementation-specific file (e.g. append a suffix) so that naming becomes unique.

crjc:
Not completely related, but would it be ok to rename:
`OgreMain/include/Animation/OgreTagPoint.h`
to
`OgreMain/include/Animation/OgreTagPoint2.h`

This seems to be the convention used for other files and would resolve conflicting names with the file:
`OgreMain/include/OgreTagPoint.h`

eugenegff:
This would be incompatible with building Ogre for multiple architectures using multiarch project in XCode or Visual Studio. Current scheme allows to enable SIMD on CMake side, but later skipping SIMD flavors that are incompatible with current architecture of multiarch project, so__OGRE_HAVE_SSE and __OGRE_HAVE_NEON macro should not trust CMake blindly, and should still check for arch compatibility.

OTOH, I see value in avoiding identically named files in source tree, will apply that part of changes

eugenegff:
So, naming changes are merged, but OGRE_SIMD_SSE2/OGRE_SIMD_NEON/OGRE_SIMD part is not, as current scheme is more flexible and allows multi architecture Ogre builds from the single source tree and single generated OgreBuildSettings.h file.

PR #169, opened 2020-12-31
Add support to set transparency sort policy per-camera

TaaTT4:
See: https://forums.ogre3d.org/viewtopic.php?p=549934#p549934

I set "sort by euclidean distance, without considering the bounding sphere radius" as a default policy. In my opinion, it's the best sorting policy for 3D scenes with perspective cameras (the main OGRE usage scenario). In fact, the same policy is used as a default by other engines as well. If this choice is too strict or monocratic, I can add a static `Camera::setDefaultSortMode` method to set the default sorting policy for cameras created from now on (similar to what has been done for `MovableObject` and `Renderable`).

### DIFF
--- a/OgreMain/include/OgreCamera.h
+++ b/OgreMain/include/OgreCamera.h
@@ -123,6 +123,13 @@ namespace Ogre {
 
         };
     protected:
+        typedef ArrayReal SortFunc( const Camera *camera, ObjectData objData );
+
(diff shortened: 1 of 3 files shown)


darksylinc:
Implemented manually to maximize performance concerns (it's a hot spot function).

However this PR was very relevant to understand what was required to do both in code and from a user perspective. Thanks!!!

PR #181, opened 2021-02-14
Make sure Ogre::Bone has same debug requirements

makuto:
This is to match `Ogre::Node`, otherwise compilation errors can result
from mismatched `OGRE_DEBUG_MODE`. In my case, I was building with `CMAKE_BUILD_TYPE=Debug`.

I don't know how to remove my merge and previous pull request commits, so I would appreciate some help if possible to edit this.


makuto:
To clarify, only 5c5c52f needs to be in this pull.

makuto:
#182 has the cleaned-up commits. Closing this one because it has unnecessary commits.

PR #182, opened 2021-02-14
 Make sure Ogre::Bone has same debug requirements

makuto:
This is to match Ogre::Node, otherwise compilation errors can result
from mismatched OGRE_DEBUG_MODE. In my case, I was building with
CMAKE_BUILD_TYPE=Debug.

This is a cleaned-up version of PR #181


darksylinc:
Thanks!!! And sorry I took so long to merge!

makuto:
No problem, thank you!

PR #192, opened 2021-04-02
Clear coat

TaaTT4:
The implementation is based on [Filament](https://google.github.io/filament/Filament.md.html#materialsystem/clearcoatmodel), taking inspiration from @rujialiu [preliminary work](https://forums.ogre3d.org/viewtopic.php?p=542452#p542452).
image
image
Differently from what Filament allows, I didn't add support neither for a clear coat layer [dedicated normal map](https://google.github.io/filament/Materials.html#materialmodels/litmodel/clearcoatnormal) nor for [base layer specular color alteration](https://google.github.io/filament/Materials.html#materialdefinitions/materialblock/lighting:clearcoatiorchange). Those two features weren't in my needs, but should be quite straightforward to add.
Another _subtle_ thing that I didn't take in account is to discriminate the clear coat lighting contribution among [standard scene rendering pass](https://github.com/google/filament/blob/main/shaders/src/light_indirect.fs#L347) and [IBL rendering pass](https://github.com/google/filament/blob/main/shaders/src/light_indirect.fs#L343).
For having a look at Filament implementation I suggest starting from [here](https://github.com/google/filament/blob/main/shaders/src/shading_model_standard.fs#L126) for direct light contribution and from [here](https://github.com/google/filament/blob/main/shaders/src/light_indirect.fs#L543) for indirect lighting contribution.

### DIFF
--- a/Components/Hlms/Pbs/include/OgreHlmsPbs.h
+++ b/Components/Hlms/Pbs/include/OgreHlmsPbs.h
@@ -601,6 +601,7 @@ namespace Ogre
         static const IdString BrdfBlinnPhong;
         static const IdString FresnelSeparateDiffuse;
         static const IdString GgxHeightCorrelated;
+        static const IdString ClearCoat;
         static const IdString LegacyMathBrdf;
         static const IdString RoughnessIsShininess;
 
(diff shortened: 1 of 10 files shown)


darksylinc:
This is something very minor:

Can I ask you that the **shader** files use tabs instead of spaces?
Right now the additional code uses both tabs and spaces.

TaaTT4:
> Can I ask you that the **shader** files use tabs instead of spaces?
> Right now the additional code uses both tabs and spaces.

Done. Sorry, I wasn't aware there was a style convention (I've seen a lot of fighting between tabs and spaces, especially in BRDF piece file) and the code taken from Filament was using spaces.

PR #200, opened 2021-05-14
Bugfix OgrePsoCacheHelper

ronaldta:
Bugfix, previously the existing equal entry was not always correctly found and new entries were created over and over again.

### DIFF
--- a/OgreMain/src/OgrePsoCacheHelper.cpp
+++ b/OgreMain/src/OgrePsoCacheHelper.cpp
@@ -96,18 +96,17 @@ namespace Ogre
     {
         RenderableCacheEntry entry;
         entry.psoRenderableKey = mCurrentState;
-        RenderableCacheEntryVec::iterator itor = std::lower_bound( mRenderableCache.begin(),
-                                                                   mRenderableCache.end(), entry );
-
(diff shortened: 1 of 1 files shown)


darksylinc:
There's a chance this bug was fixed by a combination of https://github.com/OGRECave/ogre-next/commit/055ed8f6daa2259297087372026e1faa05d2386c and https://github.com/OGRECave/ogre-next/commit/a4de4120fecc51be91e7ae8d7122a62003b4da4d

Are you able to check?

ronaldta:
Hello, [a4de412](https://github.com/OGRECave/ogre-next/commit/a4de4120fecc51be91e7ae8d7122a62003b4da4d) solves this issue. Can you please also apply it to v2-1?

darksylinc:
[Done](https://github.com/OGRECave/ogre-next/commit/50d9e1f0ebd4cde1a17b17e83cbd6a229c3db60d)!

For the record, we didn't integrate your PR because it essentially would've turned our (previously broken/buggy) binary search log N into a linear search N.

When called M times it becomes M log N vs M * N, and with a sufficiently large M the linear search becomes quadratic N²; which is really, really bad for performance.

PR #207, opened 2021-07-01
Fix alpha test check is wrongly expressed

TaaTT4:
The alpha test check is not compliant with the [_common_ specification](https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glAlphaFunc.xml). From GL specification, the formula is: _pass if the incoming alpha value is [comparison_function] to the reference value_.
Anyway, alpha test check in OGRE is bugged: eg if _equal_ is used as comparison function, the pixel is discarded (while it should pass).


TaaTT4:
Closing as WONTMERGE.
Speaking with @darksylinc, we agreed that merging this PR could (potentially) break old codebases. In addition, OGRE is obliged to follow OpenGL convention.

PR #216, opened 2021-08-02
[GL3+] add support for currentGLContext for macOS

srmainwaring:
This PR adds support for the parameter 'currentGLContext' on macOS for the GL3+ RS

There are also a couple of minor changes to help with debugging and multi-threaded applications:

  - Initialise pointers to nil in CocoaWindow::create
  - Add checks to ensure that attempts to create new Cocoa windows or set a view on an OpenGL context only occur on the main thread

I've applied the PR to v2-1 as there are difficulties compiling master on macOS at present. Ideally this change would be present on master and backported to v2-1 and v2-2. It is a pre-requisite for macOS support in the downstream project ignition-gui and ignition-gazebo.


darksylinc:
Hi!

Thanks for the PR!

I've fixed all build errors I could see when using the XCode generator on the v2-2 branch (except for MorphAnimations sample. Just comment that sample from CMakeLists)

I haven't merged those changes into master yet, but if you're in a rush hopefully master should merge with no conflicts with v2-2

PR #223, opened 2021-09-21
Fix Tutorial00 on macOS

jwwalker:
Sample_Tutorial00_Basic was hanging without displaying anything past the configuration dialog when run on macOS with Metal.  These changes add enough event handling so that it doesn't look hung and it can be quit by clicking the window close box.

### DIFF
--- a/OgreMain/include/OSX/macUtils.h
+++ b/OgreMain/include/OSX/macUtils.h
@@ -40,4 +40,7 @@ namespace Ogre {
     _OgreExport String macFrameworksPath();
     _OgreExport String macCachePath();
     _OgreExport String macTempFileName(void);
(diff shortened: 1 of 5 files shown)


darksylinc:
Niiiice!!! Thank you!

I'll be giving it a try in the next few days, if all goes ok I'll merge it

darksylinc:
Thanks for your PR!!

The only thing that could be improved is that `writeAccessFolder = /Users/user/Library/Logs` may be fine for Ogre.log but it is not for Ogre.cfg, since that's where we are supposed to store our config (that persists after multiple launches).

jwwalker:
I see; for `writeAccessFolder`, a better choice might be `~/Library/Application Support/Ogre` .

PR #225, opened 2021-09-26
[OSX] Fix segmentation fault when creating a CocoaWindow

srmainwaring:
This PR fixes a segmentation fault when creating an `Ogre::CocoaWindow` if the parameter `externalWindowHandle` is null.

**Main change**

The branch of the code creating a `CocoaWindow` with `externalWindowHandle = 0` is failing because the initialiser for the `OgreGL3PlusView` in `CocoaWindow::create` calls `Ogre::Window::getWidth()`, which in turn delegates to a contained `TextureGPU`. However this has not been initialised at the point the `OgreGL3PlusView` is created resulting in a uninitialised pointer dereference. The proposed change removes this initialiser and creates the `NSView` from a frame and then sets the Ogre::Window.

**Other changes:**

- Remove the call to `create()` from the `Ogre::CocoaWindow` constructor as it relies on the virtual behaviour of `Ogre::Window` / `Ogre::CocoaWindow`
- Subclass `Ogre::CocoaContext` from `GeneralAllocatedObject` (restoring consistency with Ogre2.1)

### DIFF
--- a/RenderSystems/GL3Plus/include/windowing/OSX/OgreOSXCocoaContext.h
+++ b/RenderSystems/GL3Plus/include/windowing/OSX/OgreOSXCocoaContext.h
@@ -33,7 +33,7 @@ THE SOFTWARE.
 
 namespace Ogre {
 
-    class _OgrePrivate CocoaContext : public GL3PlusContext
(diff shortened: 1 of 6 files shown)


darksylinc:
Thank you for your PR!

PR #229, opened 2021-09-29
fix misaligned pointer use in hash code

jwwalker:
The MurmurHash code was using misaligned pointers, which is undefined behavior according to the C++ standard and probably unsafe on multiple platforms.  See [Is it well-defined to hold a misaligned pointer, as long as you don't ever dereference it?](https://stackoverflow.com/questions/51203570/is-it-well-defined-to-hold-a-misaligned-pointer-as-long-as-you-dont-ever-deref) on Stack Overflow.

### DIFF
--- a/OgreMain/src/Hash/MurmurHash3.cpp
+++ b/OgreMain/src/Hash/MurmurHash3.cpp
@@ -11,6 +11,7 @@
 
 #include "OgrePlatform.h"
 #include "Hash/MurmurHash3.h"
+#include <string.h>
 
 //-----------------------------------------------------------------------------
 // Platform-specific functions and macros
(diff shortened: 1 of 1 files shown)


darksylinc:
My biggest concern is performance. MurmurHash was considered for its speed. This patch:

 - Replaces a raw cast with a copy (via memcpy) that may or may not be optimized (likely not, due to the visibility constraints)
 - Replaces uint32_t/uint64_t ptrs for uint8_t which has a different performance profile due to strict aliasing rules

Sounds like we should analyze the feasibility of feeding aligned-ptrs-only instead:

 1. malloc allocations are already guaranteed to be aligned (unless offseted of course)
 2. Stack allocations are aligned to whatever the data is
 3. Data coming from structures should have the proper `OGRE_ALIGNED_DECL` 
 4. Using `OGRE_ASSERT_HIGH( isAligned( key ) );` is completely acceptable
 5. Using compiler extensions to diagnose alignment is also fine

I tried using `OGRE_ASSERT_HIGH( isAligned( key ) );` with:
```cpp
inline bool isAligned( const uint8_t *x )
{
return (reinterpret_cast<size_t>(x) & 0xFu) == 0u;
}```

and it did not trigger; but a simple "works on my machine" isn't a scientific measurement.

If it's unfeasible to feed aligned ptrs, then we have no other choice but to bite the bullet

jwwalker:
To give a specific example, I'm getting odd addresses passed to the hash function by `IdString( const std::string &string )`.

eugenegff:
That is a real bug in MurmurHash3 code. It is hidden by the fact that x86/x64/arm64 allows unaligned reads, but should explode on arm32 and maybe other platforms. Other problem is with aliasing - casting (uint8*) to (uint32*) is already UB. Moreover, similar problem was found with storing results in *(uint32_t*)out = h1;

Others found this issue too, https://discourse.julialang.org/t/problem-with-issue-murmurhash3-has-undefined-behavior/9807 and I highly recommend to read that thread with justification of using memcpy compiler intrinsic to solve both issues - alignment and aliasing. Here is their fix https://github.com/JuliaLang/julia/pull/30220/files

Unfortunately, problems are not fixed in upstream https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp
and last changes to sources were made 9 years ago.

darksylinc:
Using a different hash algorithm is not out of the question.

There have been newer hashing algos since Murmur came out.

PR #230, opened 2021-10-04
Fix the condition in getClosestBlockDim2d().

mahiuchun:


darksylinc:
Well spotted!

Thank you!

Issue #232, opened 2021-10-08
Proposal: make the project side-by-side installable with Ogre (1.x)


j-rivero:
## Motivation

It would be great to be able to install this software project in a system that already has installed the previous `ogre 1.x`. Given that both projects are being developed and maintained there are some use cases where someone wants to have both installed to use them in a third party software. This is the case of the library [ignition-rendering](https://github.com/ignitionrobotics/ign-rendering) which has support for both.

We, at Open Robotics, are currently generating a Debian/Ubuntu package that is being installed together with `ogre-1.x` . To do so we are using `-DOGRE_LIB_DIRECTORY` to inject `lib/$(DEB_HOST_MULTIARCH)/OGRE-$(OGRE_VERSION)` and our packaging [scripts are moving software pieces](https://github.com/ignition-forks/ogre-2.2-release/blob/master/debian/rules#L82-L104 ) into a "versioned" directory (`OGRE-2.2`) and modifying pkgconfig/cmake to be compatible with this change.

I believe that these changes would be best hosted in the buildsystem of ogre-next so from source installations can also be installed in the same way.

Please note that I’m not proposing the changes needed to have a third party project being able to compile and link against both ogre-1.x and ogre-next at the same time, which would require some very intrusive changes in different levels.

## Changes required

Thinking on the changes needed to make the project side-by-side installable with Ogre 1.x:

 * Change project name from Ogre to Ogre-Next or Ogre-next.
 * OGRE_LIB_DIRECTORY probably needs to add an `/ogre-next/` subdirectory by default to avoid conflicting with Ogre 1.x libraries since library names are the same. Alternatively/additionally the libraries could be renamed from libOgre to libOgreNext.
 * Headers installed under typically under include/OGRE (many hardcoded paths in the code) would need to be placed in include/OGRE-Next”
 * OGRE_MEDIA_PATH from "share/OGRE/Media" to “share/OGRE-Next/Media”
 * Change the name of cmake/pkg-config modules to include `ogre-next` instead of just `ogre`. 

I'm probably missing some pieces here. Just brainstorming a little bit what I was able to found. Please feel free to correct or add the missing parts.

### For third party software using CMake/pkg-config with a system installation

The changes in pkgconfig and CMake modules should probably be enough to hide the modified paths and allow the consumers of the project to keep things working. 

[...]

PR #234, opened 2021-10-13
Change project naming to OGRE-Next

j-rivero:
Part of #232

The PR tries to start the renaming of source code in the project to match the new name `OGRE-Next`. Let me know if the project needs a different name with uppercase for 'NEXT' or plain 'next'.

Most of the changes in this PR should be harmless since they change a good bunch of comments but there a couple of commits that could potentially modify builds, other software building on top or other procedures:

 * 57b5ba changes the `project` calls from CMake that used OGRE before.
 * 594ba9c changed the `CPACK_PACKAGE_NAME` which could affect to releasing or other procedures

d5d0d3c touches README a little bit but probably it needs a major update afterwards.


darksylinc:
I wonder if the name change broke the CI.

I'll have to check it out on an actual Windows machine

PR #237, opened 2021-10-19
Add override for getCustomAttribute in MetalTextureGpu

srmainwaring:
This PR implements an override for the function `getCustomAttribute` in the class `MetalTextureGpu`.

This function is may be used to access some of the render systems underlying objects. For Metal we are interested in getting the object id for a texture. The two named attributes that may be accessed are *msFinalTextureBuffer* and *msMsaaTextureBuffer*. In both cases the object retrieved supports the `MTLTexture` protocol. The function should be called from a objective-c as follows:

```c++
id<MTLTexture> id;
texture->getCustomAttribute("msFinalTextureBuffer" ,(void*)&id);
``` 

`id` is a strong reference to the texture. The `.mm` file containing the code should be compiled with ARC enabled (in which case it is not necessary to retain or release the object as ARC will handle the reference count).


srmainwaring:
@darksylinc thanks for the feedback - I've modified the PR and force pushed the change.

PR #244, opened 2021-11-10
Examples documentation

oldmanauz:
Added a page in the manual that briefly documents all the Ogre Next samples. Also linked to it on the readme file so that people have easy access to Ogre Next features.

PR #233, opened 2021-10-12
Manual docs edits - HLMS block reference guides

oldmanauz:
Start of a PR to add reference documentation for all of the HLMS blocks. Not sure I understand how the manual links work so this commit is just to make sure I'm on the right track.

PR #246, opened 2021-11-12
Documentation refactor

oldmanauz:
This PR was originally just supposed to update the Samples documentation as per comments in #244. Some mild refactoring happened, hopefully nothing too outrageous. Changes:
- Made "Samples" it's own item in the documentation tree view
- Update samples.md with the changes from #244
- Added Tutorial_Memory to samples.md
- Modified /docs/CMakeLists.txt so that the treeview item order could be manual set
- Doxygen config file was named "html.cfg". This isn't very informative to it's function, so I renamed it to "doxygen.cfg". CMake files updated
-  Docs previously were built to "api" folder which I didn't think was overy expressive. I renamed it to "APIDocumentation". *(For reference, changes in (docs/CMakeLists.txt & doxygen.cfg.in)*
- Added headings to the Manual contents page
- Renamed some of the titles of pages in the manual to better describe what is in them
- Renamed some of the *.md files to improve consistency.
- Cleared all but one warning in the doxygen.log when building documents. Last remaining warning is to do with "graph_legend.dot" but couldn't work that one out.

Note:  #233 will conflict with this. I'll finalise that PR so you can merge that one, then merge this after.


oldmanauz:
So I think renaming the doxygen html.cfg file caused the checks to fail. Revert this particular change?

darksylinc:
We have a script [doxygen_remove_files.py](https://github.com/OGRECave/ogre-next/blob/master/.github/workflows/doxygen_remove_files.py) that failed.

Its purpose is to remove all the junk produced before uploading to ogrecave.github.io/ogre-next/api/
Getting that script right was a pita because we can't repro it locally

Maybe it's best to revert that one for now, plus it might conflict with the name change PR

darksylinc:
This PR renames "What's new in Ogre 2.2" into "What's new in Ogre 2.**3**".

I specifically don't want want to do that. The "What's new" should be preserved for historical reasons.

RIght now we have 3 "What's new pages". One for 2.2, another for 2.3 and another for 2.4

I don't have a problem "What's new" pages get moved into their own subpage for browsing historic records, but they should be kept separate and without "reversioning".

Though the latest changes (i.e. right now it's What's new in 2.4) should be on the main page for easy access.

oldmanauz:
@darksylinc 
1.
> This PR renames "What's new in Ogre 2.2" into "What's new in Ogre 2.3".

I have reverted this change in my PR. I want to tidy up the "Whats New" section but want this PR merged first so it doesn't get out of control.

2. Just to confirm, are you happy with the documentation output path change to ".../APIDocumentation" ?

darksylinc:
> Just to confirm, are you happy with the documentation output path change to ".../APIDocumentation" ?

It's probably going to raise a few headaches but I've also been bothered by the fact that there is a folder named "api" and it has nothing to do with code or interfaces at all. It's just documentation.

oldmanauz:
@darksylinc The one head ache I could think of is it might have issues with the script that updates the documentation at [ogrecave.github.io/ogre-next/api/latest](https://ogrecave.github.io/ogre-next/api/latest). 

Is there a better name than "APIDocumentation"?

oldmanauz:
@darksylinc can I nudge you on this please?

oldmanauz:
@darksylinc Can I nudge you on this please? If your not interested, let me know and I'll close it out.

darksylinc:
Apologies for taking so long.

Now that I have a fresh head, I understand what was annoying me about this PR.

This PR moves many files too much. The problem is that it will break a lot of stuff:

- Github Pages will likely start failing at first and will need fixing (fixable)
   - Changing `api` for `APIDocumentation` is a no-go.
   - If you want to go ahead with this one, this file needs to be altered: https://github.com/OGRECave/ogre-next/blob/master/.github/workflows/doxygen_remove_files.py
- External Links to docs will be broken
   - We should try to avoid breaking existing documentation unless there's a good reason (e.g. the interface changed completely).
   - Doxygen by its very nature tends to unfortunately break links (because `random%%GenErated_name` in Doxygen version 123 becomes `random__GenErated+name` in Doxygen 456. But we should avoid doing it on purpose.
   - Moving Samples to its own folder is acceptable because this section is very new. There aren't many external links (if any) to it yet
   - `manual` cannot be renamed to `Manual`

But there are many other changes that are welcome:

 - Typo fixes
 - Syntax fixes
 - The new heading organization in `manual.md`
 - Heading changes (e.g. `#` becoming `##`)

These changes are welcomed but are more controversial:

 - Changes in `Samples.md`
 - Moving samples' png files to its own folder

Therefore I suggest splitting the "welcomed non-controversial" changes into its own PR (which can be accepted immediately) and put leave the controversial changes in this PR (or a new one).

PR #254, opened 2021-12-12
Add comment to configure build to use distro packages

resttime:
Simple PR, the build instructions already suggest installing the packages from the Linux distros for use already so it's unnecessary to use the ones bundled in `ogre-next-deps`

Speeds up overall compilation and makes use of upstream packages provided by distro

### DIFF
--- a/Docs/src/SettingUpOgre/SettingUpOgreLinux.md
+++ b/Docs/src/SettingUpOgre/SettingUpOgreLinux.md
@@ -26,6 +26,8 @@ cd Ogre/Dependencies
 mkdir build
 cd build
 cmake ../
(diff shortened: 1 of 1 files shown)


darksylinc:
Thanks.

It's the neverending battle: FOSS software will always prefer using distro-provided binaries (smaller binaries, *.so loaded can be shared on ram by multiple processes, no duplicating the same libs on disk for every app, faster init time, etc).

User-compiled deps on the other hand are more useful when distributing to any OS, such as when distributing via Steam (assuming you bundle the compiled dependencies), which is basically what a Windows exe would do (i.e. package FreeImage.dll and zlib.dll alongside the main exe)

PR #255, opened 2022-01-03
Sm lights branching

dyunchik:
SM lights branching technique decreases possible shader variants for point and spot shadow-casting lights. The technique is very suitable when you have many regular (point or spot) shadow-casting lights in the scene and their combination may change in different scene locations. Default inline technique requires new shaders for every possible shadow-casting lights combination.
SM lights branching technique creates two separate loops in PS shader - one for point shadow-casting lights and one for spot lights. Make sure your lights are ordered and grouped correctly(point lights first, then - spots) if you have your own lights management.
Current implementation works for Metal, DirectX and Vulkan render systems. Default inline technique is used for OpenGL/OpenGLES.

To enable SM lights branching technique use Ogre::HlmsPbs::setMaxShadowMapLights() method before any PBS materials creation.
I've modified Sample_ShadowMapFromCode for demo and debug purposes. Use USE_STATIC_BRANCHING_FOR_SHADOWMAP_LIGHTS macro to turn on/off the technique.

### DIFF
--- a/Components/Hlms/Pbs/include/OgreHlmsPbs.h
+++ b/Components/Hlms/Pbs/include/OgreHlmsPbs.h
@@ -270,6 +270,8 @@ namespace Ogre
         void postCommandBufferExecution( CommandBuffer *commandBuffer ) override;
         void frameEnded() override;
 
+        void setMaxShadowMapLights( uint16 maxShadowMapLights ) override;
+
         /** By default we see the reflection textures' mipmaps and store the largest one we found.
             By calling resetIblSpecMipmap; you can reset this process thus if a reflection texture
             with a large number of mipmaps was removed, these textures can be reevaluated
--- a/Components/Hlms/Pbs/include/OgreHlmsPbs.h
+++ b/Components/Hlms/Pbs/include/OgreHlmsPbs.h
@@ -270,7 +270,7 @@ namespace Ogre
         void postCommandBufferExecution( CommandBuffer *commandBuffer ) override;
         void frameEnded() override;
 
-        void setMaxShadowMapLights( uint16 maxShadowMapLights ) override;
+        void setStaticBranchingLights( bool staticBranchingLights ) override;
 
         /** By default we see the reflection textures' mipmaps and store the largest one we found.
             By calling resetIblSpecMipmap; you can reset this process thus if a reflection texture
--- a/Components/Hlms/Pbs/src/OgreHlmsPbs.cpp
+++ b/Components/Hlms/Pbs/src/OgreHlmsPbs.cpp
@@ -1732,6 +1732,8 @@ namespace Ogre
 
         bool isPssmBlend = getProperty( HlmsBaseProp::PssmBlend ) != 0;
         bool isPssmFade = getProperty( HlmsBaseProp::PssmFade ) != 0;
+        bool isStaticBranchShadowMapLights =
+            getProperty( HlmsBaseProp::StaticBranchShadowMapLights ) != 0;
 
         bool isShadowCastingPointLight = false;
 
@@ -1851,6 +1853,12 @@ namespace Ogre
             {
                 // float pssmFadePoint.
                 mapSize += 4;
+            }
+            if( isStaticBranchShadowMapLights )
+            {
+                // float numShadowMapPointLights;
+                // float numShadowMapSpotLights;
+                mapSize += 4 + 4;
             }
 
             mapSize = alignToNextMultiple<size_t>( mapSize, 16 );
@@ -2300,6 +2308,16 @@ namespace Ogre
 
                 // float pssmFadePoint
                 *passBufferPtr++ = *shadowNode->getPssmFade( 0 );
+            }
+            if( isStaticBranchShadowMapLights )
+            {
+                // float numShadowMapPointLights;
+                // float numShadowMapSpotLights;
+                *reinterpret_cast<uint32 * RESTRICT_ALIAS>( passBufferPtr ) = mRealShadowMapPointLights;
+                passBufferPtr++;
+                *reinterpret_cast<uint32 * RESTRICT_ALIAS>( passBufferPtr ) = mRealShadowMapSpotLights;
+                passBufferPtr++;
+                numPssmBlendsAndFade += 2;
             }
 
             passBufferPtr += alignToNextMultiple<size_t>( numPssmSplits + numPssmBlendsAndFade, 4 ) -
@@ -3601,6 +3619,16 @@ namespace Ogre
     {
         HlmsBufferManager::frameEnded();
         mCurrentPassBuffer = 0;
+    }
+    //-----------------------------------------------------------------------------------
+    void HlmsPbs::setMaxShadowMapLights( uint16 maxShadowMapLights )
+    {
+        if( maxShadowMapLights > 0 )
+        {
+            // Make sure we calculate light positions in pixel shaders
+            setShadowReceiversInPixelShader( true );
+        }
+        Hlms::setMaxShadowMapLights( maxShadowMapLights );
     }
     //-----------------------------------------------------------------------------------
     void HlmsPbs::resetIblSpecMipmap( uint8 numMipmaps )
--- a/Components/Hlms/Pbs/src/OgreHlmsPbs.cpp
+++ b/Components/Hlms/Pbs/src/OgreHlmsPbs.cpp
@@ -3621,14 +3621,14 @@ namespace Ogre
         mCurrentPassBuffer = 0;
     }
     //-----------------------------------------------------------------------------------
-    void HlmsPbs::setMaxShadowMapLights( uint16 maxShadowMapLights )
+    void HlmsPbs::setStaticBranchingLights( bool staticBranchingLights )
     {
-        if( maxShadowMapLights > 0 )
+        if( staticBranchingLights )
         {
             // Make sure we calculate light positions in pixel shaders
             setShadowReceiversInPixelShader( true );
         }
-        Hlms::setMaxShadowMapLights( maxShadowMapLights );
+        Hlms::setStaticBranchingLights( staticBranchingLights );
     }
     //-----------------------------------------------------------------------------------
     void HlmsPbs::resetIblSpecMipmap( uint8 numMipmaps )
--- a/OgreMain/include/OgreHlms.h
+++ b/OgreMain/include/OgreHlms.h
@@ -212,10 +212,13 @@ namespace Ogre
 
         LightGatheringMode mLightGatheringMode;
         uint16             mNumLightsLimit;
+        uint16             mNumShadowMapLightsLimit;
         uint16             mNumAreaApproxLightsLimit;
         uint16             mNumAreaLtcLightsLimit;
         uint32             mAreaLightsGlobalLightListStart;
         uint32             mRealNumDirectionalLights;
+        uint32             mRealShadowMapPointLights;
+        uint32             mRealShadowMapSpotLights;
         uint32             mRealNumAreaApproxLightsWithMask;
         uint32             mRealNumAreaApproxLights;
         uint32             mRealNumAreaLtcLights;
@@ -523,6 +526,38 @@ namespace Ogre
          */
         void   setMaxNonCasterDirectionalLights( uint16 maxLights );
         uint16 getMaxNonCasterDirectionalLights() const { return mNumLightsLimit; }
+
+        /** By default shadow-caster spot and point lights are hardcoded into shaders.
+
+            This means that if you have 8 spot/point lights and then you add a 9th one,
+            a whole new set of shaders will be created.
+            Even more if you have a combination of 3 spot and 5 point lights and the combination
+            has changed to 4 spot and 4 point lights then you'll get the next set of shaders
+
+            This setting allows you to tremendously reduce the amount of shader permutations
+            by forcing Ogre to switching to static branching with an upper limit to the max
+            number of shadow-casting spot or point lights.
+
+            See    Hlms::setAreaLightForwardSettings
+        @remarks
+            All point and spot lights must share the same hlms_shadowmap atlas
+
+            This is mostly an D3D11 / HLSL SM 5.0 restriction
+            (https://github.com/OGRECave/ogre-next/pull/255) but it may also help with
+            performance in other APIs.
+
+            If multiple atlas support is needed, using Texture2DArrays may be a good solution,
+            although it is currently untested and may need additional fixes to get it working
+
+        @param maxShadowMapLights
+            Maximum number of shadow-caster spot and point lights.
+            0 to allow unlimited number of lights, at the cost of shader recompilations
+            when spot or point  lights are added or removed or their combination are changed.
+
+            Default value is 0.
+         */
+        virtual void setMaxShadowMapLights( uint16 maxShadowMapLights );
+        uint16       getMaxShadowMapLights( void ) const { return mNumShadowMapLightsLimit; }
 
         /** Area lights use regular Forward.
         @param areaLightsApproxLimit
@@ -888,6 +923,7 @@ namespace Ogre
         static const IdString DualParaboloidMapping;
         static const IdString InstancedStereo;
         static const IdString StaticBranchLights;
+        static const IdString StaticBranchShadowMapLights;
         static const IdString NumShadowMapLights;
         static const IdString NumShadowMapTextures;
         static const IdString PssmSplits;
--- a/OgreMain/include/OgreHlms.h
+++ b/OgreMain/include/OgreHlms.h
@@ -211,8 +211,8 @@ namespace Ogre
         HlmsManager *mHlmsManager;
 
         LightGatheringMode mLightGatheringMode;
+        bool               mStaticBranchingLights;
         uint16             mNumLightsLimit;
-        uint16             mNumShadowMapLightsLimit;
         uint16             mNumAreaApproxLightsLimit;
         uint16             mNumAreaLtcLightsLimit;
         uint32             mAreaLightsGlobalLightListStart;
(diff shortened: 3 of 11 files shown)


darksylinc:
Great works! Thanks!

Two things:

1. Don't worry about the merge conflicts. I'll see what I can do about that (our warning fixes + C++11-ification + clang format pretty much broke everything merging 2.3 -> master)
2. Wouldn't it be better to use the array samplers on all APIs? I see on OpenGL and Vulkan you went for:
```
	layout( ogre_t@value(texShadowMap0) ) uniform @insertpiece( TEXTURE2DSHADOW ) texShadowMap[@value(hlms_num_shadow_map_textures)];
@else
	@foreach( hlms_num_shadow_map_textures, n )
		uniform @insertpiece( TEXTURE2DSHADOW ) texShadowMap@n;@end
@end
```
But on HLSL and Metal you kept the `Texture2D<float> texShadowMap@n`. IMO it would be more consistent if all APIs follow the same path; btw the Metal syntax for arrays is a bit awkward: `array< texture2d<float>, num_textures > texShadowMap [[texture(slot)]]`.

dyunchik:
> 2. Wouldn't it be better to use the array samplers on all APIs? 

I thought about making more consistent shaders. But, I'm not sure about efficiency using the same approach with TEXTURE2DSHADOW arrays for HLSL and Metal. When I get this solution integrated in our project I'll be able to test real performance on huge scene. Approach for glslvk uses two array lookups: first gets texture index from shadowmap index and second gets real sampler from texture index. Approach for HLSL and Metal uses just one array lookup to get TEXTURE2DSHADOW.  So, I think we can start with current implementation and then test on real huge scenes both approaches.

darksylinc:
OHHHHH know I understand how you approach it! You solved the problem with an extra level of indirection!

AFAIK when it comes to shaders both approaches _should_ give the same level of performance since they should theoretically compile to the same code (either unrolled or not).

darksylinc:
I've pushed a few fixes:

 - ShadowMapFromCode was broken when `USE_STATIC_BRANCHING_FOR_SHADOWMAP_LIGHTS` wasn't defined
 - All APIs are now using the same path. I didn't get the unroll error you mention
    - And... now I get it. I'll try to fix it.
 - I fixed the array syntax which was unreadable. Fortunately we had macros to deal with it :)

There is an error I've only seen it manifesting on (Linux) Vulkan yet: very infrequently at random ShadowMapFromCode will render differently (not everything is green and the pattern changes on multiple runs, when it's broken).

I'm not sure what's causing it, I suspect either an uninitialized variable or std::sort running on pointers and the order of lights is causing it.

dyunchik:
>   * And... now I get it. I'll try to fix it.

I suspect we use old shader model, probably SM2.0 for 9.x compatibility.
we can try this pragma:
`#pragma exclude_renderers d3d11_9x`

dyunchik:
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/specifying-compiler-targets?redirectedfrom=MSDN

When you use the *_4_0_level_9_x HLSL shader profiles, you implicitly use of the Shader Model 2.x profiles to support Direct3D 9 capable hardware. Shader Model 2.x profiles support more limited flow control behavior than the Shader Model 4.x and later profiles.

darksylinc:
I've been trying to fix the "forced to unroll" without success (we may just be forced to use your approach on D3D11).

> also we can suggest compiler unroll
> @property( syntax == hlsl )[unroll(@value(hlms_lights_spot)-@value(hlms_lights_directional_non_caster))]@end 

For the unroll to work we'd need to provide an upper bound e.g. `for( ; light_idx<sm_point_lights_end || light_idx < @value( hlms_lights_point ); light_idx++ )` but it doesn't seem to be working (maybe I need to combine it with unroll like you said).

But I also stumbled with the issue that hlms_lights_point is unset to avoid variants from exploding.

> #pragma exclude_renderers d3d11_9x

I tried the pragma but got:

> warning X3568: 'exclude_renderers' : unknown pragma ignored

Anyway, we _should_ be compiling for PS 5.0 unless the FL is lower.

My main confusion is that it does work fine on WINE. The thing is, WINE uses the MS compiler (e.g. `d3dcompiler_47.dll` and co.), so if it fails on Windows it should be failing to compile on Wine, unless our the exes are linking against different compiler versions

darksylinc:
AH!!!!!!! Mystery solved!!!

It works on Release, fails on Debug. I suspect after the optimizer, the compiler is able to realize it doesn't need to unroll the loop.

dyunchik:
Indeed, hlms_lights_point was unset to avoid shader variants explosion. So we can't use it in this technique.

dyunchik:
> It works on Release, fails on Debug. I suspect after the optimizer, the compiler is able to realize it doesn't need to unroll the loop.

Anyway, we should be able to use it in Debug as well

darksylinc:
I took a look at the generated assembly.

The optimizer is just able to realize it can remove the indirection and replace it with `texShadowMap[1]` for all point & spot lights, which is why it works.

dyunchik:
> I took a look at the generated assembly.
> 
> The optimizer is just able to realize it can remove the indirection and replace it with `texShadowMap[1]` for all point & spot lights, which is why it works.

The code should work when texShadowMap[0] is used for point & spot as well.
I've changed shadownode setup for testing purposes. But generally there may be more than two shadow map textures.

darksylinc:
> The code should work when texShadowMap[0] is used for point & spot as well.
> I've changed shadownode setup for testing purposes. But generally there may be more than two shadow map textures.

Yes I know, I was just speaking out loud on why it was randomly working.

Which got me thinking if your alternative version was working because even debugger was able to realize that. So I decided to add an extra shadow map inserting a `shadowParam.atlasId = 2;` on the 3rd light and now I got a new compiler error:

```
Vulkan GLSL compiler error in 100000011PixelShader_ps:
ERROR: 0:1218: 'hlms_shadowmap3_uv_length' : undeclared identifier 
ERROR: 0:1219: '' : compilation terminated 
ERROR: 2 compilation errors.  No code generated.
```

I'm still into it

dyunchik:
I've just got the latest changes and got crash on Metal:

`Metal SL Compiler Error in 100000003PixelShader_ps:
program_source:957:4: error: expected ')'
                        array<depth2d<float>, 2>
   ^`

I can fix it, I'm on Mac right now

darksylinc:
OK just as I suspected: I tried your version of doing (in HLSL):

```hlsl
const Texture2D<float> shadowmapIdx[7] =
	OGRE_ARRAY_START( Texture2D<float> )
		hlms_shadowmap0
		
		, hlms_shadowmap1
		, hlms_shadowmap2
		, hlms_shadowmap3
		, hlms_shadowmap4
		, hlms_shadowmap5
		, hlms_shadowmap6
	OGRE_ARRAY_END;
```

This works on Debug & Release BUT turns out the compiler is just being smarter and doesn't need the optimizer to realize point & spot always use `texShadowMap[1]`

If I add one more atlas:
```cpp
			// Third light, directional, spot or point
#ifdef USE_STATIC_BRANCHING_FOR_SHADOWMAP_LIGHTS
			shadowParam.atlasId = 2; // ----> New
#endif
#ifdef USE_STATIC_BRANCHING_FOR_SHADOWMAP_LIGHTS
            shadowParam.atlasStart[0].y = 2048u;
#else
			shadowParam.atlasStart[0].y = 2048u + 1024u + 2048u;
#endif
            shadowParams.push_back( shadowParam );
```

The unroll errors come back.

dyunchik:
I fixed syntax error in Metal shader

darksylinc:
I'll be taking a short break.

I reviewed [my notes](https://forums.ogre3d.org/viewtopic.php?p=551568#p551568), it's clear we need to force-unroll on D3D11.

If there's up to 40 lights; we need to unroll 40 point and 40 spot.

We can mitigate by putting an upper limit: Up to 20 point and up to 20 spot. Thus we can unroll 20 point and 20 point, while having 40 lights.

dyunchik:
Ok, we can add loop upper constant limit

dyunchik:
Or even better:
` && cur_shadow_map<@value(hlms_num_shadow_map_lights)`

Buy I'm not sure it'll help

darksylinc:
OK it looks like FXC can't unroll because apparently `cur_shadow_map` is a divergency. I've manually unrolled the code and it still complains!

This can be compiled with `fxc.exe /T ps_5_0 /O3 /E main Test.hlsl`:

```hlsl
Texture2D<float> myTex[8];
SamplerState mySampler;

uniform int maxValues;

struct PSInput
{
	float4 color : COLOR;
};

float4 main(PSInput input) : SV_TARGET
{
	float4 retVal = float4(0,0,0,0);

	int i=0;
	int cur_shadow_map = 0;
	if( i < maxValues )
	{
		retVal += myTex[cur_shadow_map].Sample( mySampler, input.color.xy ); // OK
		++cur_shadow_map;
	}
	++i;
	if( i < maxValues )
	{
		retVal += myTex[cur_shadow_map].Sample( mySampler, input.color.xy ); //error X3512: sampler array index must be a literal expression
		++cur_shadow_map;
	}

	return retVal;
}
```

It looks like FXC didn't like that previous "++cur_shadow_map;" happening inside the branch.
If we move cur_shadow_map outside:

```hlsl
	int i=0;
	int cur_shadow_map = 0;
	if( i < maxValues )
	{
		retVal += myTex[cur_shadow_map].Sample( mySampler, input.color.xy ); // OK
	}
	++cur_shadow_map;
	++i;
	if( i < maxValues )
	{
		retVal += myTex[cur_shadow_map].Sample( mySampler, input.color.xy ); //OK
		++cur_shadow_map;
	}
```

Now it works.

This is a valid solution (because if i < maxValues is false, then i + 1 < maxValues is definitely false).
But the problem is when we later evaluate spotlights:

```hlsl
	// Point lights
	int i=0;
	int cur_shadow_map = 0;
	if( i < maxValues )
	{
		retVal += myTex[cur_shadow_map].Sample( mySampler, input.color.xy ); // OK
	}
	++cur_shadow_map;
	++i;
	if( i < maxValues )
	{
		retVal += myTex[cur_shadow_map].Sample( mySampler, input.color.xy ); //OK
		++cur_shadow_map;
	}

	// Now spotlights
	cur_shadow_map = ....;	// Umm what goes here? This is dynamic, but DX11 asks the value to be
				// a literal; it can't be passed as a uniform
	if( i < maxValues )
	{
		retVal += myTex[cur_shadow_map].Sample( mySampler, input.color.xy );
	}
	++cur_shadow_map;
	++i;
	if( i < maxValues )
	{
		retVal += myTex[cur_shadow_map].Sample( mySampler, input.color.xy ); //OK
		++cur_shadow_map;
	}
```

The only solutions to this problem I can see is:

1. Add restrictions. All pointlights must use the same atlas, all spotlights must use the same atlas. The atlas being used by spot can be different from point ones.
2. Restrict shadow maps to one type. Example 20 shadow maps exclusively for point and 20 exclusively for spot. If user has 10 point lights and 30 spot lights, only 20 spot lights can be rendered correctly

With the 2nd option the code would be like this:
```hlsl
// Point lights
int cur_shadow_map = 0;
for( int i=0; i<8; ++i )
{
	if( i < maxValues )
	{
		retVal += myTex[cur_shadow_map].Sample( mySampler, input.color.xy );
	}
	++cur_shadow_map;
}

// Spot lights
cur_shadow_map = 4; // Spotlight start
for( int i=0; i<8; ++i )
{
	if( i < maxValues )
	{
		retVal += myTex[cur_shadow_map].Sample( mySampler, input.color.xy );
	}
	++cur_shadow_map;
}
```

Which will be considered valid code by FXC.

dyunchik:
I think we can stay with one atlas for points and spots. Modern graphics cards allow to have very huge textures up to 16Kx16K. If we have 2Kx2K per light then we can support up to 64 shadow-casting lights - this is enough big amount. So we can relay on texShadowMap[0] for Directional light and texShadowMap[1] for points and spots

darksylinc:
That's true! The only issue I can see is static shadow maps, as it's often better to separate them into two atlases (because we can only clear the whole atlas, partial clears are often inefficient).

But then I remembered we support texture2darrays! I don't think we tested shadowmaps on texture2darrays so it may need fixing a few minor bugs. But other than that it should work.

If we add this restriction it may also be worth wondering: do we even need to place the shadow map in an array of textures? (Likely not)

dyunchik:
We use separate atlases for dir light and regular(point/spot) lights in our project Live Home 3D. And we have custom shadow node creation and light management. To clear each sm-rect in the  regular lights atlas we use quad pass with trivial custom material that clears depth: 
`                    if(needClearWholeAtlas)
                    {//we clear whole atlas only first one - for directional light, all other atlases serve for static spot & point lights
                        CompositorPassDef *passDef = targetDef->addPass( PASS_CLEAR );
                        CompositorPassClearDef *passClear = static_cast<CompositorPassClearDef*>( passDef );
                        passClear->setAllClearColours(clearColour);
                        passClear->mClearDepth = 1.0f;
                        passClear->mShadowMapIdx = shadowMapIdx;
                        passClear->mShadowMapFullViewport = atlasId==0;
                    }else if(needSeparateClear)
                    {//clear the rect in the atlas with special quad pass
                        CompositorPassDef *passDef = targetDef->addPass( PASS_QUAD );
                        CompositorPassQuadDef *passClearQuad = static_cast<CompositorPassQuadDef*>( passDef );
                        passClearQuad->setAllLoadActions( LoadAction::Load );
                        passClearQuad->setAllClearColours( clearColour );
                        passClearQuad->mClearDepth = 1.0;
                        passClearQuad->mMaterialIsHlms = false;
                        passClearQuad->mMaterialName = "Ogre/Depth/Clear";
                        passClearQuad->mShadowMapIdx = shadowMapIdx;
                        passClearQuad->mShadowMapFullViewport = false;
                    }
`
So, our project is really ready for two atlases :)

dyunchik:
> If we add this restriction it may also be worth wondering: do we even need to place the shadow map in an array of textures? (Likely not)

Sure we can avoid using arrays. I think it would be helpful to inform shader, using new properties, which texShadowMap we have to use for point&spot shadow casting lights. This will give us more freedom. For example if we stay with one texShadowMap0 we can pass 0, and pass 1 to use texShadowMap1, etc. So, the only limit - all points&spot lights should use one texture. 
Approach with texture2darrays will require more development and I don't know if we can render to texture in texture2darrays to get depth.

dyunchik:
Aha, I see your latest commit - nice job! Ignore my previous comment :)

dyunchik:
I think this final solution is even much better. It simply does not allow to grab too much vram for shadow maps and solves main problem - reduces shaders variants ;)

darksylinc:
I've merged your branch with master.

It now has no conflicts :)

There is one more thing I see: `setMaxShadowMapLights` asks for a uint16 but it is only ever used as a boolean i.e. `if( getMaxShadowMapLights() > 0 )`

I think this makes sense once we realize it's just a boolean settings and the max number of lights is determined by the active shadow node.

My only question is whether we are correctly handling everything i.e. what happens if:

- The shadow node supports 5 lights but only 4 or less are used
- The shadow node supports 5 lights but 6 or more are used
- Lights are added/removed from the scene at runtime

dyunchik:
> I've merged your branch with master.
> 
> It now has no conflicts :)

Great! Tomorrow I'll get it in our Live Home 3D project and I'll test the technique in real use cases. 

> There is one more thing I see: `setMaxShadowMapLights` asks for a uint16 but it is only ever used as a boolean i.e. `if( getMaxShadowMapLights() > 0 )`
> 
> I think this makes sense once we realize it's just a boolean settings and the max number of lights is determined by the active shadow node.

I reserved MaxShadowMapLights as uint16 by intuition for possible future use.

> My only question is whether we are correctly handling everything i.e. what happens if:
> 
> * The shadow node supports 5 lights but only 4 or less are used
> * The shadow node supports 5 lights but 6 or more are used
> * Lights are added/removed from the scene at runtime

I think the technique is quite safe. We collect lights by iterating through active shadow maps. So in first case(The shadow node supports 5 lights but only 4 or less are used) we'll get just another one shader with 4 lights only. The second case will produce 5 shadow casting-lights and 6th light will be skipped. 
The third case should be handled by client custom code to manage lights adding/removing. For example in our Live Home 3D project we track static lights attach/detach and assign them to shadow maps using setLightFixedToShadowMap(). non static lights are managed by Ogre, I don't see any problem there.

darksylinc:
> I think the technique is quite safe. We collect lights by iterating through active shadow maps. So in first case(The shadow node supports 5 lights but only 4 or less are used) we'll get just another one shader with 4 lights only. The second case will produce 5 shadow casting-lights and 6th light will be skipped.
> The third case should be handled by client custom code to manage lights adding/removing. For example in our Live Home 3D project we track static lights attach/detach and assign them to shadow maps using setLightFixedToShadowMap(). non static lights are managed by Ogre, I don't see any problem there.

I see! So your main problem was the combinatorial explosion caused by point & spot lights changing even though the max number of lights stayed the same.

Yes, you could bring it down even more if Hlms always produces max N lights (and avoids i.e. crashing due to nullptr lights), but as long as you're ok. Great!



> I reserved MaxShadowMapLights as uint16 by intuition for possible future use.

I had to revert that to a boolean to avoid confusion and [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

I actually thought about it: The max number of lights is strongly tied with the active shadow node. Thus that value should be auto-calculated (cached in the shadow node?) when starting the pass.

This is different from `setMaxNonCasterDirectionalLights` because there could be an arbitrary number of directional lights created regardless of the active shadow node.

Anyway, your PR has been merged.

dyunchik:
Thank you!

darksylinc:
Btw I ran into the Vulkan issue again and I was able to [fix it](https://github.com/OGRECave/ogre-next/commit/64b2d24aa24e72454e4effc647bdeabcc92facf1)

dyunchik:
> Btw I ran into the Vulkan issue again and I was able to [fix it](https://github.com/OGRECave/ogre-next/commit/64b2d24aa24e72454e4effc647bdeabcc92facf1). Update: The fix was [incomplete](https://github.com/OGRECave/ogre-next/commit/70f247eef660e2a17f2389cb966fbe47cbbfc344). Now it's done.

Great!
Is that what you saw earlier?

PR #260, opened 2022-01-12
In macCachePath, make sure our cache directory exists

jwwalker:
The macCachePath function was returning a path to a directory that had never been created, so Ogre's debug writing of shaders to the default path silently failed.


darksylinc:
Thanks for your contribution!

PR #261, opened 2022-01-18
missed code from sm_lights_branching

dyunchik:
This code was missed somewhere when sm_lights_branching was merged into the Ogre master.
Indeed it's very important not to set hlms_lights_spotparams or set it as numPoints+numSpots to decrease total shader variants


darksylinc:
Thanks!

I'm closing this PR because I could not recognize these parameters and after examination, they belong to the dead Mobile paths!

They have been [removed entirely](https://github.com/OGRECave/ogre-next/commit/83b097eb11f3c95ce59ddaa1bccde272450789e4). Thanks for pointing this out!

dyunchik:
That's even better! I was not sure if we can remove them.

PR #265, opened 2022-02-10
Python API via SWIG (WIP)

rpaciorek:
Hi,

this is proposition to add Python API to Ogre-next generated by SWIG.

The SWIG as Ogre Component approach is inspired by Python API from Ogre 1.x. This is updated (and reworked) version of my PR for 2.1 from bitbucket (abandoned with 2.1 and bitbucket).

The main question is if we want the Python API in Ogre itself or maybe as separated project (it use only headers and linking to OgreMain.so)?

API can be used to manipulate Ogre objects from python script system in C++ (where most of the Ogre related stuff is made in C++, main use case for me) or to init and use Ogre from pure Python (see [testRun.py](https://github.com/OGRECave/ogre-next/compare/master...rpaciorek:python?expand=1#diff-a515e4096e289ef6c1bb1b00b3d7a71bf3152c125e436a57a24fce5681667985) example).

Currently API is working, but incomplete and need some cleanup too (e.g. move testRun.py to samples and use Media from Ogre in it).

While creating and testing API I found some problem in Ogre itself (see FIXME notes in [Ogre.i](https://github.com/OGRECave/ogre-next/compare/master...rpaciorek:python?expand=1#diff-d0b1ca1c90e2bd8dfea69df5cca7872e31ea4a129ee0194e35be49db803199c3) - mostly protected types used in public interface and missing reference by declared but not define functions), I can create PR to fix this if you want.

I use HLMS init code from samples as Python API element (see [initHLMS.i](https://github.com/OGRECave/ogre-next/compare/master...rpaciorek:python?expand=1#diff-7d9e1e8c1443035fb665f881a4106efafdedcbf2018a1b3003dea87a7a45e5e5)), maybe Ogre should provide such helper function?


AllianceDarkSylinc:
I fixed all the `FIXME is not defined` + `MovableObject::calculateCameraDistance`.
I also just fixed `FIXME ScheduledTasks`

darksylinc:
I fixed all the `FIXME is not defined` + `MovableObject::calculateCameraDistance`.
I also just fixed `FIXME ScheduledTasks`

rpaciorek:
Thanks. You probably missed protected `typedef CameraList` vs public `getCameras()` problem in `OgreSceneManager.h`.

Any comments about idea to add Python API via SWIG?

PR #270, opened 2022-03-12
Support for UMA on Metal

dyunchik:
I've implemented support for UMA on Metal with a few lines of code. I've created a new buffer type and new VboFlag: BT_DEFAULT_SHARED and CPU_ACCESSIBLE_SHARED. These allow us use shared verteх and index buffers without any shadow buffers or meshes on UMA Apple archs(iOS and Arm macOS).


darksylinc:
Overall it's ok (except for that `else if`).

It doesn't look like there's any mechanism for synchronization so I have a few remarks:

 - It looks like it's meant to mostly replace `BT_IMMUTABLE` (or `BT_DEFAULT` as long as the GPU doesn't write to it or you manually track hazards). Documentation should make this more explicit; as currently the docs makes it look like it can easily replace `BT_DEFAULT` (any case) and `BT_DYNAMIC_*` which isn't intended. I can reword this after accepting the PR. I just need confirmation.
 - What happens if `BT_DEFAULT_SHARED` is used in a NUMA arch is untested and probably cause crashes. We should probably warn this. Again, I can add this to documentation, I just need confirmation (maybe you tested it?).

dyunchik:
Yes, you're right we have to document it better. I was hurry, I don't have much time for work because I'm in Ukraine right now, here is a war as you may know. I've moved my family in more safe place on the west, but we have а couple air raid alerts every day and every night. 
BT_DEFAULT_SHARED - is strongly recommended on GPUs with UMA, but it should slowly work on Metal even on non UMA GPUs. I haven't test it on non UMA GPUs, because I do not have Intel Mac in my hands right now. I'll ask someone from my team to test it.
Nevertheless we should recommend to use it only under RSC_UMA caps.

dyunchik:
Oh, I'll also add OS version check in RSC_UMA detection code, because api available starting from macOS 10.15

dyunchik:
My team programmer has tested BT_DEFAULT_SHARED buffers on non UMA GPU! So, now I can confirm it works on Metal on non UMA GPUs.

darksylinc:
I added fixes to your branch. Once the CI is done I'll merge your PR.

> Yes, you're right we have to document it better. I was hurry, I don't have much time for work because I'm in Ukraine right now, here is a war as you may know. I've moved my family in more safe place on the west, but we have а couple air raid alerts every day and every night.

I knew you were from Ukraine which is why I offered to make those fixes myself; but I am surprised you're still in Ukraine working! That's insanely brave.

I hope you stay well, and I'm glad your family was able to move to a safe place.

dyunchik:
I can't move abroad, because I'm 47 and in theory I may fight for my motherland. Eugene Golushkov was able to move to Poland in first day of Russian invasion. Thanks Covid-19 we can work remotely ;) and thanks God I've moved and stay with my family in Khmelnitsky city far way from the war zone. Unfortunately, Russian military may attack any part of county with cruise missiles. That's why sometimes we have air raid alerts.

dyunchik:
Nevertheless, we have to work to be able to support our country and our army!

PR #275, opened 2022-03-24
Correct docs of Light::setRectSize

jwwalker:
By experimental results, Light::setRectSize sets the full size, not the half size, of an area light.


darksylinc:
I took a look and you're right. I was sure it was otherwise.

PR #277, opened 2022-04-13
Make Frustum::getProjectionParamsAB work with orthographic projection

jwwalker:
Frustum::getProjectionParamsAB was assuming a perspective projection.  This change makes it work with orthographic projection too, and documents it accordingly.


jwwalker:
It doesn't look like the failure of the Linux check is my fault.

darksylinc:
Yeah, the ABI checker script is not meant to be run yet. If the CI made it to the ABI checker phase, build was a success.

The logic to tweak when the ABI checker should be run needs to be tweaked.

PR #283, opened 2022-04-28
fix for upload to BT_DEFAULT_SHARED buffers - we have to be sure the last command buffer's completed

dyunchik:
BT_DEFAULT_SHARED were designed for rare uploads. Nevertheless if we need to upload new data we have to be sure that last committed command buffer's completed to avoid artefacts due to data changing on the go. Indeed on some heavy scenes gpu may still execute last committed command buffer that uses BT_DEFAULT_SHARED vertex/index buffers when we come to upload some new data. Of cause we can wait for previous frame completion or even use stall, but these techniques will switch current command buffer that may be unacceptable. That's why I've made a new technique waitUntilCommitedCommandBufferCompleted() that waits for the last committed command buffer completion instead of the last frame completion with command buffer switching, so we can continue to work with current command buffer.


dyunchik:
Oops, looks like I touched ABI

dyunchik:
Do I have to change minor version to overcome ABI checker ?

darksylinc:
Hi! The ABI checker should not be running yet, it's broken. It should start running once 3.0 is frozen and we start with 4.0 development (so that any change done to 3.0 branch doesn't break ABI).

We need see why [the logic](https://github.com/OGRECave/ogre-next/blob/92285e0713e01277eea6a7f54d5212dc9ecffe3c/.github/workflows/linux.build.yml#L24) is broken and fix it.

As for your PR, my biggest concern is that this change is incredibly:

 1. Specific to your needs
 2. Specific to Metal
 3. `VaoManager::waitUntilCommitedCommandBufferCompleted` is completely undocumented. It probably won't do nothing in Vulkan which will be misleading and cause errors.

I may be more willing to accept this PR if it's renamed to `_waitUntilCommitedCommandBufferCompleted` to indicate it's advanced/internal use and if some documentation is added on what it does, why is it needed, and that currently it only works on Metal.

dyunchik:
This technique to wait until the last committed cmd buffer's completed can be implemented on Vulkan as well using vkQueueWaitIdle(). But you're right, right now it's really specific to Metal. So renaming to _waitUntilCommitedCommandBufferCompleted() is a good idea.

dyunchik:
the main idea of this technique is not to commit current cmd buffer, user can continue with current cmd buffer. We just need to be sure that previously committed commands have already done and gpu does not use our BT_DEFAULT_SHARED data.

darksylinc:
Because it was rebased, this PR popped into my inbox.

Looking at it again, I have 2 questions:

1. Does Metal work fine if you never use `BT_DEFAULT_SHARED` at all? I see the code holds on to a `mCommitedCommandBuffer` reference and it's only reset in the next commit or if `waitUntilCommitedCommandBufferCompleted()` is called (which would be if you use `BT_DEFAULT_SHARED`). I'm asking this question because I vaguely remember that holding on to the command buffer after issuing had some side effects.
2. I see that the code is basically doing a hammer and is almost always a stall. Because it always waits for the last issued command. In other words it looks like it's (**almost!**) functionally equivalent to `vaoManager->waitFor( vaoManager->getFrameCount() );`. Does it "just work" if you just try calling that instead of `waitUntilCommitedCommandBufferCompleted()`?
    - I say almost because the waitFor() will also close the currently open command encoder which in this case doesn't seem necessary and may be undesirable depending on when is the wait happening.

dyunchik:
Hi Matias!

We use BT_DEFAULT_SHARED on platforms with support for UMA (iOS, macOS Apple silicon, macOS with Intel embedded GPU). Metal works with BT_DEFAULT as well, but it's better to use BT_DEFAULT_SHARED for UMA. 
Usually we don't need to change buffers with BT_DEFAULT_SHARED. But if we need to change them, we have to be sure that gpu does not use them.

vaoManager->waitFor( vaoManager->getFrameCount() ) does almost the same effect as waitUntilCommitedCommandBufferCompleted(), but it closes current command buffer! So waitUntilCommitedCommandBufferCompleted()(we call it only if we upload some new or changed data) is more accurate to make sure that previously committed commands(executed right now on GPU) do not use buffer that is going to be changed.
The same problem, should exist on Vulkan as well.

dyunchik:
Imagine that commands that are going to current command buffer should operate with upcoming new data in uploading buffer. If we do waitFor() before data upload these commands will operate with old data. Usually we don't do that way, we prepare data(vertices and indexes) before adding commands, but I'm not sure if that isn't possible.

Another reason: waitFor() calls stall() that leads to more Obj-C Metal calls and sends _notifyDeviceStalled() with possible side effects.

darksylinc:
> Imagine that commands that are going to current command buffer should operate with upcoming new data in uploading buffer. If we do waitFor(vaoManager->getFrameCount()) before data upload these commands will operate with old data. Usually we don't do that way, we prepare data(vertices and indexes) before adding commands, but I'm not sure if that isn't possible.

While that is possible using raw Vulkan and Metal, it is not really "allowed" with OgreNext because D3D11 support mandates that data must be uploaded before issuing commands.

In OgreNext and using Metal or Vulkan, any command involving data transfers will likely open the copy encoder (also called "blit encoder") which means the previous encoders will be closed; and that may or may not mean that commands will committed to the GPU.
Meaning that if you issued commands and upload data later, you may end up with wrong render in OgreNext.

> Another reason: waitFor(vaoManager->getFrameCount()) calls stall() that leads to more Obj-C Metal calls and sends _notifyDeviceStalled() with possible side effects.

Yes, though most of that code is meant to ensure the rest of OgreNext knows that the device has been stalled; which means all operations don't need to wait at all (and memory waiting to be released can be released ASAP).
If you do a stall without notifying OgreNext, nothing bad happens but the underlying interfaces still think the CPU is behind the GPU.

Overall I have mixed feelings about this PR:

1. In one hand, it's not the way I would do things (regarding UMA memory support). With more work, it could be arranged in a way that CPU & GPU can still work in parallel (for the most part).
2. On the other hand, I don't have the time to work on it; and your interface is a highway to address a specific (but relevant to mobile w/ low memory and slow CPU/GPU) problem to edit buffers while consuming low memory. The "only" cost are some stalls here and there.

I think the original point stands: Just rename waitUntilCommitedCommandBufferCompleted to _waitUntilCommitedCommandBufferCompleted and it's good for inclusion.

dyunchik:
Ok, I'll add docs and rename it to _waitUntilCommitedCommandBufferCompleted()

dyunchik:
Done that

dyunchik:
Ops, looks like my changes don't conform to clang-format style 😕
Checker log is useless. I don't have any tool on my MacBook to reformat code sources. Have to search for some third party clang-format XCode extensions ☹️
That makes contributing from XCode quite awkward. Clang-format requirements may significantly decrease further contributions from users, especially those who use XCode as a main IDE.

darksylinc:
Don't worry about the clang format failure, I can fix it myself.

If you're able to run:

```bash
cd .github/workflows
python3 run_clang_format.py --fix_broken_files
```

Then great, if not then I'll do it.

dyunchik:
I'll try run_clang_format script. Thank you!

dyunchik:
It's a nightmare. I've installed homebrew and then installed clang-format, fixed clang-format-13 link to clang-format, ran the script and finally it has not helped! I've still got Clang-Format Check failure! Less files, but still failure.

darksylinc:
I ran the script again and that just fixed the issues.

Either the version of clang-format wasn't the same, it had to be ran twice (it happens...), or it didn't show up in git to commit.

Anyway, I fixed it myself and merged.

Thanks for the PR!

PR #286, opened 2022-04-30
When a doxygen comment applies to a preceding member or method, it ne…

jwwalker:
…eds to start with ///<, not just ///.  Otherwise it applies to the wrong thing.


darksylinc:
Do you know if clang format was run afterwards?
Some of these changes look like they might trip the columns per line limit, causing clang-format to breakup the line

jwwalker:
@darksylinc  I assume clang-format was not run afterward, because I don't know anything about that or how to run it.  Can you point me in the right direction?

darksylinc:
Hi!

Some IDEs have it builtin automatically hence why I asked.

The "normal" way of running it is by installing clang format 13 (that's what we use; [how to install](https://releases.llvm.org/download.html) it depends on which OS you are on).

So you would do:

`clang-format-13 myFile.cpp`

And it will modify `myFile.cpp` applying the formatting style.

darksylinc:
- Visual Studio has it builtin and will ask with a pop up when it detects `.clang-format` configuration file (which we do provide)
 - QtCreator also has it in the Options -> Beautifier -> Clang Format

jwwalker:
I'm normally working on a Mac using Xcode, not Visual Studio or QtCreator.  I guess I'll do it with a command line.

jwwalker:
OK, I added another commit with changes made by clang-format 13.

PR #293, opened 2022-05-12
Fix some Mac-specific deprecation warnings.

jwwalker:
Most of these changes are just replacing a constant defined in Apple's headers with a newer one that does the same thing.


jwwalker:
By the way, none of these newer names are newer than macOS 10.12, so it's not like this change is going to break things for someone who is not quite up to the latest and greatest.

PR #298, opened 2022-06-05
Jw/doxygen misc warnings

jwwalker:
This fixes doxygen warnings of several kinds:

- Unsupported xml/html tag
- documented empty return type of
- unexpected end of comment block while parsing the argument of


jwwalker:
@darksylinc Thanks, I have made that fix and other improvements to the docs of the `Cmov4` and `CmovRobust` methods.

PR #311, opened 2022-07-13
Fixed bug in "OgreColourInterpolatorAffector" incorrectly in first frame

ronaldta:
Fixed bug in "OgreColourInterpolatorAffector", which is not at all applied for the first frame after particle creation. Bug is visible at lower framerates specially when fading-in a particle.


darksylinc:
I fixed the minor issues for you.

Thanks for your contribution!

PR #317, opened 2022-07-30
Add parameter to skip material assignment to Items on creation

36yang:
When creating an item, add a parameter to specify whether to use the associated material in mesh.

Because the default material will load all textures, we can avoid loading useless textures when we want to set other materials. Especially in my editor, I need to create icons for hundreds of models under the resource folder.

All I need is a blackwhite material. But the default mesh material loads all textures. As a result, icon creation time is very long.

### DIFF
--- a/OgreMain/include/OgreAny.h
+++ b/OgreMain/include/OgreAny.h
@@ -150,7 +150,7 @@ namespace Ogre
                 return OGRE_NEW_T( holder, MEMCATEGORY_GENERAL )( held );
             }
 
-            void writeToStream( std::ostream &o ) override { o << held; }
+            void writeToStream( std::ostream &o ) override {  }
 
         public:  // representation
             ValueType held;
--- a/OgreMain/include/OgreAny.h
+++ b/OgreMain/include/OgreAny.h
@@ -150,7 +150,7 @@ namespace Ogre
                 return OGRE_NEW_T( holder, MEMCATEGORY_GENERAL )( held );
             }
 
-            void writeToStream( std::ostream &o ) override {  }
+            void writeToStream( std::ostream &o ) override { o << held; }
 
         public:  // representation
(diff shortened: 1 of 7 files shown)


darksylinc:
Hi!

Let me know about the `void writeToStream( std::ostream &o ) override` change.

Once that's done I can take care of the clang format issues and merge your PR.

36yang:
The std::ostream has no implent of << ， It only to  resolve compile problem there.

darksylinc:
Thanks for your contribution!

I removed the `void writeToStream( std::ostream &o ) override { o << held; }` change from this PR.

I think the error you saw was caused due to `#include <ostream>` not being included, so I fixed that already.

However if that's not the problem it could be that the type needs to implement the `<<` operator.

Anyway, the more I kept looking at OgreAny.h the more eyebrows I kept raising. The whole class could use being written from start. It's heavily bloated.

Hence if it needs work I decided it would be better to do that separately.

jwwalker:
> Anyway, the more I kept looking at OgreAny.h the more eyebrows I kept raising. The whole class could use being written from start. It's heavily bloated.

Just curious, what is the advantage of `Ogre::Any` over `std::any`?

darksylinc:
That code was written more than 10 years ago, long before std::any existed.

First Boost::Any existed, which is what std::any is based on; but adding Boost as a dependency just for that was overkill.

Anyway, Boost is over-engineered (std::any is as well) specially for this purpose; so it's not surprising Ogre::Any is as well.

By over-engineered I meant it hogs compile times, adds runtime performance to debug builds; for what is essentially a glorified type-safe void\* / uint64.

There are cases where it makes sense, but in this context an uint64_t is more than enough (if the user needs more, it can store more data into a pointer then cast the ptr to an uint64_t).

PR #320, opened 2022-08-01
Remove unused variable: fAutoConstantBiasScale

mjcarroll:
The compiler flags this as an unused variable.  As it turns out, most of
the function body is a no-op, so I removed it as well.

Alternatively, there may be some additional logic that should have been here that was missed? 

### DIFF
--- a/OgreMain/src/Compositor/OgreCompositorShadowNode.cpp
+++ b/OgreMain/src/Compositor/OgreCompositorShadowNode.cpp
@@ -992,22 +992,6 @@ namespace Ogre
     {
         const ShadowTextureDefinition &shadowMapDef =
             mDefinition->mShadowMapTexDefinitions[shadowMapIdx];
-
-        float fAutoConstantBiasScale = 1.0f;
-        if( shadowMapDef.autoNormalOffsetBiasScale != 0.0f )
-        {
-            Light const *light = mShadowMapCastingLights[shadowMapDef.light].light;
-            OGRE_ASSERT_HIGH( light && "Can't call this function if isShadowMapIdxActive is false!" );
-
-            const Camera *texCamera = mShadowMapCameras[shadowMapIdx].camera;
-            if( texCamera->getProjectionType() == PT_ORTHOGRAPHIC )
-            {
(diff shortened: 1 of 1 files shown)


darksylinc:
Oh I remember this snippet.

The variable is supposed to be used; but this bug was discovered too late and would cause some of our users' bias tweaks to need tweaking again; thus we avoided fixing the bug for 2.3

It was [fixed in in 3.0](https://github.com/OGRECave/ogre-next/blob/79294827d251df49a94bb202a5ae65870bc4258e/OgreMain/src/Compositor/OgreCompositorShadowNode.cpp#L1013).

Removing this block may cause us to erroneously delete it when merged into 3.0 when resolving merge conflicts.

Do you have any suggestion to clarify and deal with this situation?

mjcarroll:
> Do you have any suggestion to clarify and deal with this situation?

I believe there are a few other unused variable warnings.  I think it may be sufficient to add a macro such as:

```
OGRE_UNUSED_VAR(x) (void)(x);
```

Would you be okay with this?  If so, where should the macro live?

jwwalker:
> I believe there are a few other unused variable warnings. I think it may be sufficient to add a macro such as:
> 
> ```
> OGRE_UNUSED_VAR(x) (void)(x);
> ```

I have a vague recollection that in some compilers, trying to deal with unused variables using `(void)` may not work or may even cause more warnings.  If OgreNext assumes C++17 (I don't recall if it does), then the right way to do it is:

```
[[maybe_unused]] float fAutoConstantBiasScale = 1.0f;
```

darksylinc:
Is there a reason this warning is special in particular? OgreNext 2.3 has **TONs** of warnings. All of them fixed in 3.0

This sounds like a whack-a-mole hunt unless I'm missing something?

mjcarroll:
> Is there a reason this warning is special in particular? OgreNext 2.3 has TONs of warnings. All of them fixed in 3.0

Nope, I was just working through my local set of patches that I have been making over time.  If it's all fixed in 3.0, then I won't spend any more effort on it.

PR #321, opened 2022-08-01
Change bitwise operations to logical

mjcarroll:
Newest versions of clang warn on this via warning: use of bitwise '|' with boolean operands [-Wbitwise-instead-of-logical]

The comments indicate that maybe this was done for performance reasons.  I think without profiling/benchmarking to prove otherwise, it probably makes sense to prefer logical operations in these cases.


mjcarroll:
Same as #320 , just sending some of my local patches.  Sounds like these are all addressed in 3.0, so not necessary to land in 2.3

PR #327, opened 2022-08-18
Unregister Windows class when destroying

mjcarroll:
This prevents issues with repeated library loads in the same process.


eugenegff:
This commit fixes scenario of D3D11 RS .dll reloading, when mClassRegistered variable is reset but WNDCLASS remained registered.
OTOH sequential creation of the second window after destroying first will fail. Also app will crash in multiwindow scenario after first window destruction.

The best way to fix it is to unregister WNDCLASS not on the first window destruction but on .dll unloading, at the time of dll's global variables destruction, when code of WNDPROC is unloaded from the process.
This way if .dll will be reloaded - it will start with cleared mClassRegistered and non-registered WNDCLASS, the same as at the first run.

PR #332, opened 2022-08-29
Fix the texture mipmap not auto generated when use unlit.

36yang:
Fix the texture mipmap not auto generated when use unlit.


darksylinc:
Hi!

I'm on vacations so I may take a few weeks to review/merge this PR.

Issue #264, opened 2022-02-10
Custom projection matrix breaks the shadow


Th3V1kt0r:
#### System Information
- Ogre Version: master (10.02.2022)
- Operating System / Platform:  Win7
- RenderSystem: OpenGL
- GPU: GeForce GTX 1060 6GB

#### Detailled description

If I use my own projection matrix, then the shadow is displayed incorrectly. 

You can also see the error if you take Sample_ShadowMapDebugging and use Ogre's own projection matrix as custom projection matrix.

For example like this:
``` 
void ShadowMapDebuggingGameState::createScene01()
{
...
   Ogre::Camera* camera = mGraphicsSystem->getCamera();
   Ogre::Matrix4 proj = camera->getProjectionMatrix();
   camera->setCustomProjectionMatrix(true, proj);
... 
}
```

Th3V1kt0r:
It seems that the shadow camera does not work properly when the projection matrix is custom. If I define the projection matrix as not custom, then the shadow looks reasonable.
Example of how it works for me:


       void PSSMShadowCameraSetup::getShadowCamera(
             ...
             const Matrix4 oldProjection = cam->getProjectionMatrix();
             const bool iscustom = cam->isCustomProjectionMatrixEnabled();
             if (iscustom)
             {
                 _cam->setCustomProjectionMatrix(false, oldProjection);
             }
             if( iteration < mNumStableSplits )
             {
                  mConcentricShadowCamera.getShadowCamera( sm, cam, light, texCam, iteration, viewportRealSize );
                  mMinDistance = mConcentricShadowCamera.getMinDistance();
                  mMaxDistance = mConcentricShadowCamera.getMaxDistance();
             }
             else
             {
                 FocusedShadowCameraSetup::getShadowCamera( sm, cam, light, texCam, iteration, viewportRealSize );
             }
             if (iscustom)
             {
                 _cam->setCustomProjectionMatrix(true, oldProjection);
             }

PR #333, opened 2022-09-03
fix custom projection

jwwalker:
When a custom projection matrix is in use, `Frustum::updateFrustumImpl` should call `renderSystem->_convertProjectionMatrix` rather than `renderSystem->_makeRsProjectionMatrix`, because the latter can overwrite customized entries of the projection matrix.


Th3V1kt0r:
Hi [jwwalker](https://github.com/jwwalker).
I found your commit to custom matrix. I have problems with the shadows when using the custom matrix. Do you use the custom matrix? Does the shadow work for you?
More about my problem: [https://github.com/OGRECave/ogre-next/issues/264](https://github.com/OGRECave/ogre-next/issues/264)

PR #338, opened 2022-10-04
Replaced deprecate functions in Win64

GermanAizek:
@darksylinc,
If you need support for new Win64 systems, these changes are necessary.


darksylinc (approved):

eugenegff (approved):
As specified in MSDN

> Note  To write code that is compatible with both 32-bit and 64-bit versions of Windows, use SetWindowLongPtr. When compiling for 32-bit Windows, SetWindowLongPtr is defined as a call to the [SetWindowLong](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-setwindowlonga) function.

so we should simply always use SetWindowLongPtr, that is #defined to SetWindowLong in non _WIN64 case.

PR #339, opened 2022-10-05
Code refactor for readability

GermanAizek:
@darksylinc minor changes affect readability code, but `clear()` function cleans string correctly

### DIFF
--- a/Components/Overlay/src/OgreOverlayElementCommands.cpp
+++ b/Components/Overlay/src/OgreOverlayElementCommands.cpp
@@ -89,7 +89,7 @@ namespace Ogre
             }
             void CmdMaterial::doSet( void *target, const String &val )
             {
-                if( val != "" )
+                if( !val.empty() )
                 {
                     static_cast<OverlayElement *>( target )->setMaterialName( val );
                 }
--- a/OgreMain/src/OgreMeshSerializerImpl.cpp
+++ b/OgreMain/src/OgreMeshSerializerImpl.cpp
@@ -1599,7 +1599,7 @@ namespace Ogre
                                                             unsigned short lodNum, MeshLodUsage &usage,
                                                             uint8 casterPass )
         {
-            usage.manualName = "";
+            usage.manualName.clear();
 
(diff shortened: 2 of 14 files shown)


darksylinc:
Although this has a moderate amount of subjectivity, not every change improves readability. In fact it can make it  worse.

For example this:

```cpp
response = "";
// or this:
binOptList["-O"] = "";
```

With no context tells me that `response` and `binOptList["-O"]` are very likely a string. And that we want to initialize them to an empty string. There's a slight chance it could be a custom class that overrides the `=` operator for strings, but this chance is very minor.

However this:


```cpp
response.clear();
// or this:
binOptList["-O"].clear();
```

Doesn't tell me anything. These variables could be a vector, a map, a list, a string, a custom class. Without an IDE and further reading I can't tell. Is `response` supposed to hold a list of responses? Binary responses?

There are cases such as in [OgreScriptLexer.cpp](https://github.com/OGRECave/ogre-next/pull/339/files#diff-2e269f371180369fcf06571d09b6ce70ed9d138218314f17072212b3fc23267c) where the change may likely result in higher performance, and it is a place where performance is important, hence the change is welcome.

Likewise in `D3D11HLSLProgram::setTarget` it is clearly obvious from context that we are dealing with strings, so `mTarget.clear();` is probably better (subjective).

Now `variable.empty()` vs `variable != ""` is far less controversial, because the information conveyed is already enough (the fact that `variable` is empty) and in most places the fact that it is a string is also inferred from context easily (e.g. `assert( ( profileName != "" ) && ( "Profile name can't be an empty string" ) );` is obvious). Additionally, `variable.empty()` covers more cases than `variable != ""` (when the string is corrupt, and avoids bizarre errors such as zero-length spaces creeping into the source code)

I can extract some of these changes selectively from your PR and merge them manually; so that we can debate the rest.

GermanAizek:
@darksylinc,
I understood about readability with ur examples and I worsened it. I don't mind if you accept some changes, I'm making an offer, and yours to accept or refuse.

PR #344, opened 2022-10-31
HLSL has no mix function, but lerp.

Th3V1kt0r:
HLSL has no mix function, but lerp.
In GLSL this is handled by a define.
Otherwise an assertion is triggered when I activate DebugPssmSplits.


darksylinc:
Thanks!

PR #357, opened 2023-01-07
IBL changes to maximize compatibility with other PBR engines

### LINKED ISSUE #190: Unwanted division by PI in reflection map contribution
First things first, let me summarize how divisions by PI are handled in OGRE (to check that my assumptions are good):
* The PI at the denominator of the normal distribution function (Trowbridge-Reitz GGX) is stripped away from the [`R` calculation](https://github.com/OGRECave/ogre-next/blob/master/Samples/Media/Hlms/Pbs/Any/Main/200.BRDFs_piece_ps.any#L138) and re-applied subsequently, in the [`G` calculation](https://github.com/OGRECave/ogre-next/blob/master/Samples/Media/Hlms/Pbs/Any/Main/200.BRDFs_piece_ps.any#L145) (geometry function);
* The division by PI in diffuse BRDF calculation ([`Rd` term](https://github.com/OGRECave/ogre-next/blob/master/Samples/Media/Hlms/Pbs/Any/Main/200.BRDFs_piece_ps.any#L174) in OGRE shader) is not present since [it's already embedded](https://github.com/OGRECave/ogre-next/blob/master/Components/Hlms/Pbs/src/OgreHlmsPbsDatablock.cpp#L489) in `kD` (which is then stored in `pixelData.diffuse`);

If everything is OK so far, I'll move on.
When it comes time to [add the reflection contribution](https://github.com/OGRECave/ogre-next/blob/master/Samples/Media/Hlms/Pbs/Any/Main/200.BRDFs_piece_ps.any#L232) to the final color, `pixelData.diffuse` is still affected by a PI division which, at this stage, shouldn't be present anymore (or at least counteracted).

darksylinc:
Fixes #190 

The division by PI was indeed unwanted; as that is something specific to lambertian diffuse integration over the hemisphere to maintain energy conservation.

However IBL is about _incoming_ light from the hemisphere, not a single point like with directional lights.

This PR also includes a lot of other fixes/improvements.


dyunchik (commented):
This is a mistake! We have to use fresnelS - it takes in account pixelData.NdotV, but pixelData.F0 does not.

darksylinc:
@dyunchik I moved the discussion to #361 since this is a closed PR.

PR #362, opened 2023-02-17
Fix Xcode compile error caused by header confusion

jwwalker:
Without this change, Xcode 14.2 gives compile errors in src/OgreTagPoint.cpp.  It seems that when it says #include "OgreTagPoint.h", it's reading Animation/OgreTagPoint.h.


darksylinc:
For some reason in CMake 3.21.1 / XCode 13.4.1 HEADERMAP = NO was the default; and thus I was not seeing this.

So yes, we should explicitly disable it to get it working.

Thanks!

PR #365, opened 2023-03-15
Avoid transitive include of ctime

nkoenig:
Adds <ctime> include to avoid it's current transitive inclusion.

### DIFF
--- a/OgreMain/src/GLX/OgreTimer.cpp
+++ b/OgreMain/src/GLX/OgreTimer.cpp
@@ -27,6 +27,7 @@ THE SOFTWARE.
 */
 #include "OgreTimer.h"
 #include <sys/time.h>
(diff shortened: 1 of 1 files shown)


darksylinc:
Hi!

Is there a particular problem this PR is fixing? Could you elaborate?

nkoenig:
This is coming from a downstream user that strictly follows the "Include what you use" principle. https://google.github.io/styleguide/cppguide.html#Include_What_You_Use

This doesn't fix any systematic problem with OGRE. Please decline this if you'd prefer.

darksylinc:
I was leaning towards rejecting this PR until your explanation.

While I don't think I agree with what cppcheck is saying here; a clean log is more important when the cost is so little here.

PR #372, opened 2023-04-18
remove old unused files and msvc references <version 1800

mosfet80:
"Other" folder contains unreferenced old software.
remove msvc references < v1800


darksylinc:
I'll have to test this by hand. Some of the warnings disables you're removing were introduced in a particular MSVC version but that doesn't mean they can be re-enabled.

Particularly I'm almost certain the CRT warning disables are still needed

mosfet80:
I re-entered the #pragma code

PR #376, opened 2023-04-22
Clean and update Andoid section

mosfet80:
remove unused old android.toolchain.cmake
update settingup Ogre Andoid to support vulkan https://developer.android.com/ndk/guides/graphics/getting-started


darksylinc:
Hi!

> Android 8.0 or newer strongly recommended. Android 7.0 and 7.1 are supported,
> but most phones are bundled with very old and buggy drivers.

This is relevant info that's being removed.

> Android 19 or newer strongly recommended. 

This is wrong info. There is no Android 19 yet (currently at 13), and if this is SDK level, Android SDK level 19 is not supported by OgreNext either.

> Android NDK is stored at /home/username/Android/Sdk/ndk/21.0.6113669
> Android NDK is stored at /home/username/Android/Sdk/ndk/XX.X.XXXXXXX

I'm unsure about this one. People prefer a working example for the moment things go sideways.

We cannot guarantee every single XX.X.XXXXXXX version works with Ogre (Google has the really bad habit of introducing breaking changes to the NDK) so using a real valid version is useful both as an example as well as a working NDK version.

>    -DANDROID_NATIVE_API_LEVEL=24 \
>    -DANDROID_NATIVE_API_LEVEL=25 \

Wrong, we support API level 24. Is there a reason to bump it to 25?

mosfet80:
i rechecked it is correct that nkd remains 24 ... 25 is the last one. Excuse me!

I removed ndk/21.0.6113669 as it doesn't support vulkan

Reading the requirements of NDK24:

Android NDK r24 (March 2022)
[Changelog](https://github.com/android/ndk/wiki/Changelog-r24)
Jelly Bean (APIs 16, 17, and 18) is no longer supported. The minimum OS supported by the NDK is **KitKat (API level 19)**.

darksylinc:
> I removed ndk/21.0.6113669 as it doesn't support vulkan

That can't be right... I'm using that NDK version and it supports Vulkan (in fact OgreNext won't work without Vulkan support on Android).

Anyway the rest LGTM, it's been merged. Thanks!

PR #380, opened 2023-04-28
Fix Compile Warning.

mosfet80:
when i compile ogre, i have this warning: OpenGL_GL_PREFERENCE has not been set to "GLVND" or "LEGACY", so for  compatibility with CMake 3.10 and below the legacy GL library will be used.
To solve the warning:
Policy CMP0072 says that I need to set OpenGL_GL_PREFERENCE to LEGACY.


darksylinc:
Thanks for the PR!

Issue #373, opened 2023-04-20
Normals not handled correctly for non-uniform scaling


jwwalker:
#### System Information
- Ogre Version: master branch [0a6cd24a91]
- Operating System / Platform: macOS 13.3.1
- RenderSystem: Metal
- GPU: AMD Radeon Pro 560X

#### Detailed description
When the model-view transformation includes a non-uniform scale, the correct way to handle normal vectors is to multiply by the inverse transpose of the upper left 3x3 part of the model-view matrix.  As far as I can tell, this is not what happens in Ogre.  In `800.VertexShader_piece_ps.any`, I see the line

```
@property( hlms_normal || hlms_qtangent )	outVs.normal	= mul( @insertpiece(local_normal), toMidf3x3( worldViewMat ) );@end
```

which seems to be just multiplying the local normal by the 3x3 part of the model-view matrix.

A visual illustration:  First, an unscaled cylinder using lit PBS shading.
![shot_00002 - Scene 1, Shot 2  unscaled, Ogre](https://user-images.githubusercontent.com/730426/233502261-55e7339f-42a7-4982-ba74-09be536f9d0e.jpg)

Here's what it looks like when the cylinder is stretched in the horizontal direction perpendicular to the camera's viewing direction:
![shot_00004 - Scene 1, Shot 4  scaled, Ogre](https://user-images.githubusercontent.com/730426/233502450-967a7114-d119-4a3d-8b7f-0b27dd1e6f47.jpg)
(Notice that the specular highlight is basically unchanged, in spite of the surface being flatter there.)

And here is what I think it ought to look like, as rendered by a different rendering engine:
![shot_00003 - Scene 1, Shot 3  stretched, classic](https://user-images.githubusercontent.com/730426/233502561-aa3ba877-e271-4ff5-a05e-686b666b6ab9.jpg)

darksylinc:
There's discussion here: https://forums.ogre3d.org/viewtopic.php?t=96656

Most people would like to simply ignore this problem due to the performance implications.

But for those who want correctness, the suggested:

```
@property( accurate_non_uniform_scaled_normals )
   normalMat = transpose( inverse( worldMat ) );
@end
```

Could be a good solution (note: IIRC Metal and HLSL don't implement `inverse()` as an intrinsic since it's expensive).

jwwalker:
[...]

PR #381, opened 2023-04-28
Add vertex shader code for accurate non-uniform-scaled normals

jwwalker:
This is to address issue #373.


darksylinc:
Can I ask you to give me write access to your fork? There's a minor style fix I want to push before merging the PR.

Thanks.

darksylinc:
I forgot to ask: Where do you think this toggle is best?

1. Per Item: Possible but a bit clunky to support & maintain from OgreNext side
2. Per Material: Easy
3. Per Pass: Also easy

The main benefit of being per material is that you can toggle it for some materials that will use Items that need it; without sacrificing the perf cost to everything else.

The main benefit of per pass is that you don't have to worry about forgetting to toggle it on for a material/Item pair that needs it.

PR #382, opened 2023-05-02
+bugfix attributes should be assigned in pairs

Th3V1kt0r:
Attributes should be assigned in pairs.
Here it was probably forgotten.

### DIFF
--- a/RenderSystems/GL3Plus/src/windowing/win32/OgreWin32GLSupport.cpp
+++ b/RenderSystems/GL3Plus/src/windowing/win32/OgreWin32GLSupport.cpp
@@ -722,8 +722,10 @@ namespace Ogre
 
 #if OGRE_NO_QUAD_BUFFER_STEREO == 0
             if( SMT_FRAME_SEQUENTIAL == mStereoMode )
+            {
                 attribList.push_back( WGL_STEREO_ARB );
-            attribList.push_back( GL_TRUE );
+                attribList.push_back( GL_TRUE );
+            }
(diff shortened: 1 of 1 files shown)


darksylinc:
Thanks!

I have a question, out of curiosity: Do you use Quad Buffer Stereo?

I ask because we still left the code "just in case", but QBS is a really old feature from NV Quadros back when the industry was pushing for 3D stereo, but today newer ways of rendering stereo has superseded this (such as the Sample InstancedStereo or the OpenVR one).

So I'm curious if someone is still actually using it.

Th3V1kt0r:
Hi. We have used QBS on OpenVR for our optional feature on Ogre1. A few years ago we tried to migrate our project to Ogre2. Unfortunately the goal was not achieved. The migrating was stopped. Maybe it will be continued again sometime. Currently we continue with Ogre1 and I sometimes merge it to Ogre2.--Diese Nachricht wurde von meinem Android Mobiltelefon mit GMX Mail gesendet.Am 02.05.23, 21:05 schrieb "Matias N. Goldberg" ***@***.***>:

   
  Thanks! 
  I have a question, out of curiosity: Do you use Quad Buffer Stereo? 
  I ask because we still left the code "just in case", but QBS is a really old feature from NV Quadros back when the industry was pushing for 3D stereo, but today newer ways of rendering stereo has superseded this (such as the Sample InstancedStereo or the OpenVR one). 
  So I'm curious if someone is still actually using it.

PR #383, opened 2023-05-04
bugfix: memory leak due to undeleted stereo driver bridge

Th3V1kt0r:
Stereo driver must be freed after device is created.

This bug is probably also included in Ogre1.

### DIFF
--- a/RenderSystems/Direct3D11/src/OgreD3D11RenderSystem.cpp
+++ b/RenderSystems/Direct3D11/src/OgreD3D11RenderSystem.cpp
@@ -206,6 +206,12 @@ namespace Ogre
             mHLSLProgramFactory = 0;
         }
 
+#if OGRE_NO_QUAD_BUFFER_STEREO == 0
+        // Stereo driver must be freed after device is created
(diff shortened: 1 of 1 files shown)


Th3V1kt0r:
When we get our product working under Ogre2, then we will look at OpenVr. At the moment I can't say if we will use QBS in the future. I uploaded it so that when we update to higher version we don't have to keep updating our bug fixes.But I promise not to upload QBS bugs.I still have a few bugfixes. But I can't see directly if they are universal or specific for us.Can I send it to you and you decide?--Diese Nachricht wurde von meinem Android Mobiltelefon mit GMX Mail gesendet.Am 05.05.23, 15:37 schrieb "Matias N. Goldberg" ***@***.***>:

   
   
   out of interest: are you actually using QUAD_BUFFER_STEREO or just cleaning-up code? 
   
  He answered that question here #382 (comment) as I had the same question

AllianceDarkSylinc:
I merged it manually to fix the clang format check fail.

Thanks for the PR!

darksylinc:
I merged it manually to fix the clang format check fail.

Thanks for the PR!

PR #390, opened 2023-05-22
Docs fixes

cryham:
I made many fixes in hlms doc.
Funny thing: the more I read the more I find them. Maybe it'd be easier to just paste this all to some spellchecking editor.

I added more `custom_*` blocks that I found in .any files, they need description, instead of `..` placeholder.

One more thing, don't know how to fix it, if you look at website
https://ogrecave.github.io/ogre-next/api/latest/hlms.html
there are few links that don't work (get linked) and odd text is there instead. Search for `|outline` e.g.: `9.3.Hlms templates|outline "collected"`, `9.4.1.5.`, `9.6.1.preparePassHash|outline`, `see 8.9.1.4.Loading a texture twice` etc.

Well I still don't know answers to my questions like here:
https://forums.ogre3d.org/viewtopic.php?p=554575#p554575
https://forums.ogre3d.org/viewtopic.php?p=554600#p554600
I think it would be good to add some (even small) section(s) with such answers or links. I think these are rather common for advanced users.

### DIFF
--- a/Docs/src/manual/hlms.md
+++ b/Docs/src/manual/hlms.md
@@ -331,6 +331,9 @@ case sensitive. The following keywords are recognized:
 -   \@insertpiece
 -   \@pset padd psub pmul pdiv pmod pmin pmax
 
+Note: @ takes precedence over comment //. If you comment out with //
+a line that has @, it will _not_ be commented out.
+
 ### \@property( expression ) {#HlmsPreprocessorSyntaxProperty}
 
 Checks whether the variables in the expression are true, if so, the text
--- a/Docs/src/manual/hlms.md
+++ b/Docs/src/manual/hlms.md
@@ -145,7 +145,7 @@ was cut as well. The HLMS default systems handle these.
     these files (or write some of their own) to fit their custom needs.
 3.  **C++ classes implementation.** The C++ takes care of picking the
     shader templates and manipulating them before compiling; and most
-    importantly it feeds the shaders with uniform/constans data and sets
+    importantly it feeds the shaders with uniform/constants data and sets
     the textures that are being in use. It is extremely flexible,
     powerful, efficient and scalable, but it's harder to use than good
     ol' Materials because those used to be data-driven: there are no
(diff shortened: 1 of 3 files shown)


Docs/src/manual/hlms.md
```diff
@@ -821,32 +823,38 @@ customized:
...
+| custom_vs_uniformStructDeclaration |  .. |
+| custom_vs_posMaterialLoad    |  .. |
+| custom_vs_preTransform       |  .. |
 | custom_vs_preExecution       |  Executed before Ogre's code from the Vertex Shader.|
 | custom_vs_posExecution       |  Executed after all code from the Vertex Shader has been performed.                     |
 | custom_ps_uniformDeclaration |  Same as custom_vs_uniformDeclaration, but for the Pixel Shader|
+| custom_ps_uniformStructDeclaration  |  .. |
 | custom_ps_preExecution       |  Executed before Ogre's code from the Pixel Shader.|
 | custom_ps_posMaterialLoad    |  Executed right after loading material data; and before anything else. May not get executed if there is no relevant material data (i.e. doesn't have normals or QTangents for lighting calculation)|
+| custom_ps_posSampleNormal    |  .. |
```
darksylinc, line 852: We can use `[TBD]` instead of `..`.

The thing about some of this customization pieces is that they were added for a particular reason; so it's more a thing of "I need to add a piece of code that happens in this *exact* spot" rather than explaining what it exactly is supposed to do; look where it is in the code, and override it.

darksylinc:
Wow, that makes me feel embarrassed and I hope my English teacher never sees this post \<deletes all traces of your PR entirely\>

Ok jokes aside. Good job!

> there are few links that don't work (get linked) and odd text is there instead. Search for |outline e.g.: 9.3.Hlms templates|outline "collected", 9.4.1.5., 9.6.1.preparePassHash|outline, see 8.9.1.4.Loading a texture twice etc.

It looks like they became broken back from when we switched from the [odt manual](https://github.com/OGRECave/ogre-next/blob/master/Docs/2.0/Ogre%202.0%20Porting%20Manual%20DRAFT.odt) to Doxygen. I just fixed [what you highlighted](https://github.com/OGRECave/ogre-next/commit/f1add1e548c2e4d501dfe2f836fa3f9ad0f164b8).

darksylinc:
Before accepting the PR I think my only grip is with this change:

> C++ can use `Hlms::setProperty( "key", value )` to set "key" to the given
> value. This value can be read by `\@property`, `@foreach`,
>
> vs
>
> C++ can use `Hlms::setProperty( "key", value )` to set value to the given
> "key". This value can be read by `\@property`, `@foreach`,

cryham:
Ah right, I just realized it's the same meaning. I restored it.

PR #396, opened 2023-07-10
Delete Tools/OgreZip2Cpp/src directory

mosfet80:
remove unused file with folders (already removed in ogre)


eugenegff:
That tool is complimentary to EmbeddedZipArchive, it is not used in Ogre, but intended to be used by users of Ogre to implement resources embedding workflow for their game. Primary usage case for EmbeddedZipArchive was NaCL port, but till EmbeddedZipArchive is still with us - tool should be too.

PR #399, opened 2023-07-13
Terra fix shader w/o metalness or roughness map(s)

cryham:
Just like I wrote here:
https://forums.ogre3d.org/viewtopic.php?p=554764#p554764
Now I can use values set by `terraDatablock->setRoughness` and `terraDatablock->setMetalness`, without maps.
BTW IDK why there's 3 commits here, previous are merged already.


cryham:
I made it again with fixed, 1 commit history in #414, so closing this.

PR #413, opened 2023-08-31
Add sortByName to HlmsJson::saveMaterials

cryham:
So by default as before. But with true, we can have sorted materials by name saved.


darksylinc:
Merged manually to get the clang format fixes.

Thank you for the PR!!!

PR #436, opened 2024-03-04
Fix OpenGL Win32 window ignoring "hidden" parameter

mixrigster:
Currently, when creating a render window on OpenGL Win32 with the "hidden" parameter set to true the window remains visible.
This is because `mHidden` is never set correctly and while the window is correctly created invisible, it immediately gets updated afterwards based on the (wrong) value of `mHidden`.

This fix sets `mHidden` to the correct value and also removes the redundant call to `SetHidden()`.


darksylinc:
This PR spent too much time unmerged. Sorry for that!

Thanks for your contribution!

PR #443, opened 2024-04-19
OgreRootLayout uses json elements despite OGRE_NO_JSON.

Th3V1kt0r:
Some code parts use json elements despite OGRE_NO_JSON. As a result, OgreMain is not compiled.


darksylinc:
LGTM.

Thanks!

PR #444, opened 2024-04-26
Diffuse/specular parameters are mixed incorrectly. Or not?

Th3V1kt0r:
Diffuse/specular parameters are mixed incorrectly. Or not?
In any case, there are visible differences.
Left picture as it is. Right picture with my changes.

I have only set ambient light in the scene: 
`sceneManager->setAmbientLight( Ogre::ColourValue( 1.0f, 0.0f, 0.0f ), Ogre::ColourValue( 1.0f, 1.0f, 1.0f ), Ogre::Vector3::UNIT_Y);`

example

### DIFF
--- a/Samples/Media/Hlms/Pbs/Any/AmbientLighting_piece_ps.any
+++ b/Samples/Media/Hlms/Pbs/Any/AmbientLighting_piece_ps.any
@@ -78,9 +78,9 @@
 			{
 		@end
 				pixelData.envColourS += lerp( midf3_c( passBuf.ambientLowerHemi.xyz ),
-											  midf3_c( passBuf.ambientUpperHemi.xyz ), ambientWD );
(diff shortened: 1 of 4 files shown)


darksylinc:
WHAT THE...?!?!

Yes this a bug.
And it's been there since like forever???

I know that the ambient hemisphere lighting is just a hack, but still...

I'll have to evaluate how much impact it makes on the scenes because it seems to be significant enough to be mentioned by the documentation.

Release 3.0 will also be making a [change to PBR](https://ogrecave.github.io/ogre-next/api/latest/_p_b_s_changes_in30.html) so it'd be great if we can include it with that release.

darksylinc:
I've finally had some time to look into this.

Due to how significant this change can be, I added the ability to switch between the old and new method.

Thank you for your PR! It's been merged by hand.

cryham:
Holy f*ing cow :eyes:

PR #448, opened 2024-05-20
Main: move the adjacency flag to render operation where it belongs

bishopnator:
instead of the geometry shader - for e.g. rendering adjacency buffers without an geometry shader attached.

patch based on
https://forums.ogre3d.org/viewtopic.php?f=4&t=94417#p542277 https://github.com/OGRECave/ogre/pull/830/commits/ae9da1ba6f869846d5e46dcc80c55d23fbdd4708


darksylinc:
This PR has been merged to master, but will not be merged against 2.3.

Thanks for your work!

PR #455, opened 2024-07-26
move the adjacency flag to render operation where it belongs ogre-next-3-0

bishopnator:
move the adjacency flag to render operation where it belongs instead of the geometry shader - for e.g. rendering adjacency buffers without an geometry shader attached.


OgreMain/include/OgreGpuProgram.h
```diff
@@ -478,13 +478,9 @@ namespace Ogre
             return mVpAndRtArrayIndexFromAnyShader;
         }
 
-        /** Sets whether this geometry program requires adjacency information
-            from the input primitives.
-        */
+        /// @deprecated
```
darksylinc, line 481: Regarding deprecation we introduced the `OGRE_DEPRECATED_VER( n )` macro where n is the major version in which it was deprecated.

Thus the code should look like this:

```cpp
OGRE_DEPRECATED_VER( 4 ) virtual bool isAdjacencyInfoRequired()
```

(master branch is currently 4)

darksylinc:
I've pushed two changes to your PR (one to fix Clang Format, another to port your enum changes to Vulkan).

Once CI is done checking I'll merge your PR.

darksylinc:
Thank you for the PR!!!

PR #468, opened 2024-10-29
Fix include path error when compiling Ogre as a subproject

wakeofluna:
Fixes a compile error when compiling Ogre as a subproject (using FetchContent) in a different master project.


darksylinc:
Thanks!

PR #473, opened 2024-11-06
Do minor changes to cmake files

Apteryks:
This is a partial port of the commit
3568f43cd5b1e246aec3f1347fc353f388885f1a from the ogre (not ogre-next) repository, omitting the zip header change which doesn't apply.

  * MacroLogFeature: Clarify that even if a package is found it will not be used if a feature is disabled
  * Use list append on CMAKE_MODULE_PATH to allow prepending it with toolchain file, which is used with conan recipe.


Apteryks:
Closed in favor of https://github.com/OGRECave/ogre-next/pull/474 (they'd conflict if submitted individually).

PR #477, opened 2024-11-16
Memoryless tiler

eugenegff:
RSC_TILER + TextureFlags::Tiler[Depth]Memoryless => MTLStorageModeMemoryless and VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT

### DIFF
--- a/OgreMain/include/OgreDepthBuffer.h
+++ b/OgreMain/include/OgreDepthBuffer.h
@@ -89,6 +89,11 @@ namespace Ogre
             POOL_NO_DEPTH = 0,
             POOL_MANUAL_USAGE OGRE_DEPRECATED_ENUM_VER( 3 ) = 0,
             POOL_DEFAULT = 1,
+
+            /// The depth buffer doesn't does not have memory backing it.
+            /// It will be created with TextureFlags::TilerMemoryless.
+            POOL_MEMORYLESS = 65533,
+
             /// Deprecated.
             ///
             /// Do NOT use this flag directly. It made sense in Ogre 2.1
--- a/OgreMain/include/OgreRenderPassDescriptor.h
+++ b/OgreMain/include/OgreRenderPassDescriptor.h
@@ -204,6 +204,10 @@ namespace Ogre
 
     protected:
         void         checkRequiresTextureFlipping();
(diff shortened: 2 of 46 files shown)


dyunchik:
First of all, we should be able to globally disable the use of memoryless. Because some very heavy scenes can give an out of tile memory error and simply not render anything (we had such cases on Metal in Live Home 3D). For this purpose we introduced "WindowMemoryless"  param. Actually it's almost the same as "memoryless_depth_buffer". But if "WindowMemoryless"/"memoryless_depth_buffer" is false, then we have to remove MTLStorageModeMemoryless from depth buffer and also from **MSAA buffer** as well. Indeed, this way Metal will be able to split rendering into several passes with intermediate results stored in these buffers, and Metal does this. I don't know about Vulcan, but I think he rules in the same way.

dyunchik:
Second, we should be able to use memoryless for depth only when we render to  MSAA texture. For this purpose we need something like TextureFlags::TilerDepthMemoryless

dyunchik:
Hm, may be the special memoryless depth pool is enough.

darksylinc:
>  Because some very heavy scenes can give an out of tile memory error and simply not render anything (we had such cases on Metal in Live Home 3D).

Oh! I didn't know this. Then yes, we need a global toggle.

The only issue I see is that such toggle requires quite a bunch of resources to be recreated (if you want it to be toggled at runtime and not at startup).

Regardless Vulkan, ARM Mali has a 180MB limit (memoryless **or not**). That 180MB is in the data that gets sent from VS to PS, i.e. position + interpolators data.

ARM [published more details about this](https://developer.arm.com/documentation/SDEN-3735689/latest) yesterday, see page page 13.

> Hm, may be the special memoryless depth pool is enough and we may avoid TilerDepthMemoryless

Yes, the memoryless depth pool should be enough.

darksylinc:
OK I just pushed the necessary change:

Added "Allow Memoryless RTT" RenderSystem option which calls `TextureGpuManager::setAllowMemoryless`.

If all is well, it can be merged.

PS: I noticed clang format script was ignoring the Vulkan folder :scream: 
We'll fix that once that is merged to avoid any potential merge conflict.

Issue #483, opened 2024-12-05
Compile error of OgreD3D11RenderSystem.cpp in 32-bit MSVC


jwwalker:
#### System Information
- Ogre Version: master branch commit [e68d10562d]
- Operating System / Platform: Windows 11
- RenderSystem: Direct3D11

#### Detailed description
I get a compile error:

```
2>D:\MoreCode\Ogre\ogre-next-fork\RenderSystems\Direct3D11\src\OgreD3D11RenderSystem.cpp(1764,65): error C2668: 'Ogre::StringConverter::toString': ambiguous call to overloaded function
2>    D:\MoreCode\Ogre\ogre-next-fork\OgreMain\include\OgreStringConverter.h(117,23):
2>    could be 'Ogre::String Ogre::StringConverter::toString(bool,bool)'
2>    D:\MoreCode\Ogre\ogre-next-fork\OgreMain\include\OgreStringConverter.h(112,23):
2>    or       'Ogre::String Ogre::StringConverter::toString(long,unsigned short,char,std::ios_base::fmtflags)'
2>    D:\MoreCode\Ogre\ogre-next-fork\OgreMain\include\OgreStringConverter.h(109,23):
2>    or       'Ogre::String Ogre::StringConverter::toString(size_t,unsigned short,char,std::ios_base::fmtflags)'
2>    D:\MoreCode\Ogre\ogre-next-fork\OgreMain\include\OgreStringConverter.h(105,23):
2>    or       'Ogre::String Ogre::StringConverter::toString(unsigned long,unsigned short,char,std::ios_base::fmtflags)'
2>    D:\MoreCode\Ogre\ogre-next-fork\OgreMain\include\OgreStringConverter.h(91,23):
2>    or       'Ogre::String Ogre::StringConverter::toString(int,unsigned short,char,std::ios_base::fmtflags)'
2>    D:\MoreCode\Ogre\ogre-next-fork\OgreMain\include\OgreStringConverter.h(75,23):
2>    or       'Ogre::String Ogre::StringConverter::toString(double,unsigned short,unsigned short,char,std::ios_base::fmtflags)'
2>    D:\MoreCode\Ogre\ogre-next-fork\OgreMain\include\OgreStringConverter.h(71,23):
2>    or       'Ogre::String Ogre::StringConverter::toString(float,unsigned short,unsigned short,char,std::ios_base::fmtflags)'
[...]

PR #485, opened 2024-12-16
Fix overload ambiguity problem wth StringConverter::toString

jwwalker:
Ensure that `StringConverter::toString` can be used without ambiguity or conversion by providing overloads for all 6 32 or 64 bit integer types.  See: https://github.com/OGRECave/ogre-next/issues/483


darksylinc:
Thanks!

PR #486, opened 2024-12-17
getMemoryStats was reporting used staging texture bytes as available, and vice versa

jwwalker:
TextureGpuManager::getMemoryStats was reporting used staging texture bytes as available, and vice versa


darksylinc:
Thanks!

PR #487, opened 2024-12-19
Fix error in LwString with template parameter declared but. not used

jwwalker:
The 8-parameter overload of LwString::a had a template parameter T that was declared but not used.


darksylinc:
Wow! This bug stayed hidden for a long time.

Thanks!

Issue #502, opened 2025-02-25
OGRE_USE_NEW_PROJECT_NAME not fully working on Mac with Xcode


jwwalker:
#### System Information
- Ogre Version: master branch commit [ca8475d510b]
- Operating System / Platform: macOS 13.7.4 
- RenderSystem: Metal
- GPU: Radeon Pro 560X 

#### Detailed description
I use the CMake GUI, and verify that `OGRE_USE_NEW_PROJECT_NAME` is checked.  I run it and have it make an Xcode project.  When I open the project, there is a target named OgreNextMain, but its `PRODUCT_NAME` build settings is just Ogre, so this target will build Ogre.framework, not OgreNext.framework.  The other libraries seem OK, e.g., the OgreNextHlmsPbs target has a `PRODUCT_NAME` value of OgreNextHlmsPbs.

PR #504, opened 2025-03-06
fix the failure to use the OgreNext name in Xcode project

jwwalker:
I think this fixes bug #502 .


darksylinc:
Thanks!

Hopefully it will!

PR #511, opened 2025-03-17
Improved support for strong macroblocks (and blendblocks).

bishopnator:
The implementation added an option to override more settings from a macroblock and a blendblock as well. It is possible to do it through inherited classes of Ogre::Hlms or Ogre::HlmsListener attached to an existing instance of Hlms. The caller is responsible to orginize the description of the updates through the active properties or it is possible to use build-in support of Hlms which uses a single property for macroblock overrides and a single property for blendblock overrides.

### DIFF
--- a/OgreMain/include/OgreHlmsPso.h
+++ b/OgreMain/include/OgreHlmsPso.h
@@ -65,7 +65,8 @@ namespace Ogre
         /// For multi-GPU support
         uint32 adapterId;
 
-        uint8 strongMacroblockBits;
+        /// @see StrongBasicBlocks enum
+        uint8 strongBasicBlocks;
 
         bool operator==( const HlmsPassPso &_r ) const
         {
@@ -78,7 +79,7 @@ namespace Ogre
                    this->depthFormat == _r.depthFormat &&              //
                    this->sampleDescription == _r.sampleDescription &&  //
                    this->adapterId == _r.adapterId &&                  //
-                   this->strongMacroblockBits == _r.strongMacroblockBits;
+                   this->strongBasicBlocks == _r.strongBasicBlocks;
         }
         bool operator!=( const HlmsPassPso &_r ) const { return !( *this == _r ); }
         bool operator<( const HlmsPassPso &other ) const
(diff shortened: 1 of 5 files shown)


darksylinc:
This change will break the ABI and thus cannot be targetted against 3.0 (at least as is)

Please select master as the target branch.

OgreMain/include/OgreHlmsPso.h
```diff
@@ -141,12 +212,22 @@ namespace Ogre
         bool          enablePrimitiveRestart;
         uint8         clipDistances;  // Bitmask. Only needed by GL.
 
+        uint8 strongBlocks; // @see StrongBlocks enum.
```
darksylinc, line 215: Doesn't this member hold the same data as HlmsPso::pass->strongBasicBlocks ?

OgreMain/src/OgreHlms.cpp
```diff
@@ -2141,51 +2145,206 @@ namespace Ogre
...
-            // Depth writes is already off, we don't need to hold a strong reference.
-            pso.pass.strongMacroblockBits &= ~HlmsPassPso::ForceDisableDepthWrites;
-        }
-        if( pso.macroblock->mCullMode == CULL_NONE )
-        {
-            // Without culling there's nothing to invert, we don't need to hold a strong reference.
-            pso.pass.strongMacroblockBits &= ~HlmsPassPso::InvertVertexWinding;
-        }
-        if( pso.macroblock->mDepthClamp )
+        assert( pso.strongBlocks = 0 );
```
darksylinc, line 2150: Use OGRE_ASSERT_LOW

OgreMain/src/OgreHlms.cpp
```diff
@@ -2141,51 +2145,206 @@ namespace Ogre
...
+
+        // Modify a macroblock if needed.
+        HlmsMacroblock macroblock = *pso.macroblock;
+        macroblock.mRsData = nullptr;
+        mListener->applyStrongMacroblockRules( macroblock,
+                                               *this );  // Allows the listener to modify a macroblock
+        applyStrongMacroblockRules(
+            macroblock );  // Allows the implementation (inherited classes) to modify a macroblock
+        assert( macroblock.mRsData == nullptr );  // Check if implementation reassigned the macroblock
```
darksylinc, line 2159: Use OGRE_ASSERT_LOW

darksylinc, line 2159: Change:
`// Check if implementation reassigned the macroblock`
for:
`// Check implementation didn't reassign the macroblock.`

OgreMain/src/OgreHlms.cpp
```diff
@@ -2499,7 +2658,7 @@ namespace Ogre
         pso.blendblock = datablock->getBlendblock( casterPass );
         pso.pass = passCache.pso.pass;
 
-        applyStrongMacroblockRules( pso );
+        applyStrongBlockRules( pso );
```
darksylinc, line 2661: I feel like applyStrongBlockRules() can be early out if pso.pass.strongBlocks is 0, signifying there is no request to change anything at all.

Previously we didn't do that because the operation was ridiculously cheap, but now it's a bit more expensive now.

OgreMain/include/OgreHlmsPso.h
bishopnator: Actually I forgot to delete HlmsPso::pass->strongBasicBlocks - this member is not used anymore.

OgreMain/src/OgreHlms.cpp
bishopnator: No, it is not so east now - the method sets the flags. There is early return inside for macroblock adjustments and blendblock adjustement. I wanted to remove this redundancy, that it is necessary to request for the update of xxxblock and also additionally set some properties. Now the caller sets the properties and update the xxxblock if needed. The Hlms checks if the xxxblock was changed and in this case sets the flag for strong ref and get the new block from HlmsManager. If you really wish, I can add the flags that a block is requested, but it looks like just a redundant information. Here it is not required to have it as fast as possible as the method is not called very often. On the other hand, I think it is not a good idea to have such a condition for calling methods from HlmsListener.

PR #514, opened 2025-03-28
Revove old Visual C++ support

mosfet80:
Clean code.
removed code referring to old versions of visual c++ that are no longer supported.


darksylinc:
Wow, some of these settings were ANCIENT! Thanks!

PR #515, opened 2025-03-29
Revove old Visual C++ support II

mosfet80:
removed code referring to old versions of visual c++ that are no longer supported.


darksylinc:
Thanks!

PR #520, opened 2025-04-20
Missing init of mTexture pointer in OgreFont.

Hotshot5000:
If calling loadImpl() it will call guiDatablock->setTexture( 0, mTexture, &samplerblock ); with unit mTexture pointer instead of NULL. In unloadImpl() it will call textureManager->destroyTexture( mTexture ); with random pointer.


darksylinc:
Thanks!

PR #528, opened 2025-06-26
Fix CompositorWorkspaceDef::connect failing with aliased nodes.

nimble0:
The `connect ( IdString outNode, IdString inNode )` overload doesn't resolve aliases before looking up the node definition.


darksylinc:
Thanks!

PR #535, opened 2025-07-28
Add loongarch64 support

zhaixiaojuan:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.

Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:
https://loongson.github.io/LoongArch-Documentation/README-EN.html

### DIFF
--- a/OgreMain/include/OgrePlatform.h
+++ b/OgreMain/include/OgrePlatform.h
@@ -60,6 +60,7 @@ THE SOFTWARE.
 #define OGRE_CPU_PPC 2
 #define OGRE_CPU_ARM 3
 #define OGRE_CPU_MIPS 4
+#define OGRE_CPU_LOONGARCH 5
 
 /* Find CPU type */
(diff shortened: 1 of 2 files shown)


OgreMain/include/OgrePlatform.h
```diff
@@ -73,6 +74,8 @@ THE SOFTWARE.
 #    define OGRE_CPU OGRE_CPU_ARM
 #elif defined( __mips__ ) || defined( __mips64 ) || defined( __mips64_ ) || defined( _M_MIPS )
 #    define OGRE_CPU OGRE_CPU_MIPS
+#elif defined( __loongarch64 )
```
xen0n: Bitness is not reflected here, so just `__loongarch__` will do.

```suggestion
#elif defined( __loongarch__ )
```

PR #542, opened 2025-09-21
Include inner dirs in OGRE-Hlms.pc

marcin-serwin:
The Hlms headers depend on the headers in Common to be passed via `-I` option, e.g., `OgreHlmsUnlitDatablock.h` includes `OgreHlmsTextureBaseClass.h` instead of `Common/OgreHlmsTextureBaseClass.h`.


darksylinc:
Thanks!

PR #543, opened 2025-10-03
fix cmake deprecation

mosfet80:
cmake version < then 3.10 is deprecated


darksylinc:
Thanks!

Though we should probably use min...max syntax next time, which doesn't need constant updates to the scripts.

PR #545, opened 2025-10-13
fix seg fault in screen space reflections

skrobchik:
Patch 1/6 of various fixes to the Vulkan backend

Originally authored by @sssarana while at Qualcomm.

Creating pull request as the open-source contribution request has been approved.

@sssarana feel free to amend any commits as they're no longer CCI.

### DIFF
--- a/Samples/Media/2.0/scripts/materials/ScreenSpaceReflections/GLSL/CopyColourAndDepth_cs.glsl
+++ b/Samples/Media/2.0/scripts/materials/ScreenSpaceReflections/GLSL/CopyColourAndDepth_cs.glsl
@@ -1,55 +1,42 @@
 #version 430
 
-uniform sampler2D srcRtt;
-@property( !texture1_msaa )
-	uniform sampler2D srcDepth;
-@end @property( texture1_msaa )
-	uniform sampler2DMS srcDepth;
-@end
+vulkan_layout(ogre_t0) uniform texture2D srcRtt;
+#if !texture1_msaa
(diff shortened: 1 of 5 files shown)


darksylinc:
Well!

With the final PR #550 in this series being merged, this concludes your series of 6 PR patches!

First, apologies for taking so long. I was working on a contract and barely had any time left to review it.

Second, **thank you very much** for this PR! Both of you, and to Qualcomm who allowed to upstream these changes.

I do have a question (it's ok if you can't answer). I had the sensation that some of these changes were used to automate testing at Qualcomm? If so, that's amazing!

We've run into Qualcomm GPU driver bugs in the past, in fact [we've got workarounds](https://github.com/OGRECave/ogre-next/blob/db27780181b646011576a299d92a0ceb6ae82a26/OgreMain/include/OgreWorkarounds.h) for Android flavor of Adreno drivers. Most notably we always disable indirect draws after we've found that Adreno 618 (but not other models like the 650) would trigger a GPU reset when calling vkCmdDrawIndexedIndirect with CbDrawIndexed::instanceCount = 0 (or was it with drawCount = 0? I forgot).

Another bug we found is that vkCmdBlitImage is broken for 3D textures (at least on old drivers like driver 512.502.0). Since we don't know when it was fixed, we always fallback to Software blitting for 3D textures on Qualcomm drivers (IIRC currently no OgreNext sample showcases 3D texture blitting though).

If you're going to be testing for Android + Qualcomm, make sure all workarounds are turned off.

And again, thanks!

PR #549, opened 2025-10-13
allow setting num of frames from command line

skrobchik:
Patch 5/6 of various fixes to the Vulkan backend

See #545


darksylinc:
Merged manually due to trivial merge conflicts with the previous PRs.

Issue #539, opened 2025-08-31
CompositorNode::notifyRecreated has exponential complexity


nimble0:
#### System Information
- Ogre Version: master
- Operating System / Platform: Linux
- RenderSystem: GL3Plus / Vulkan
- GPU: AMD Radeon RX 6600 XT

#### Detailled description
`CompositorNode::notifyRecreated` calls the same method on connected nodes. https://github.com/OGRECave/ogre-next/blob/77d40b816c65fa12332c00267fac0616961165b9/OgreMain/src/Compositor/OgreCompositorNode.cpp#L280 A simple chain of nodes with 2 or more connections between them all is sufficient to cause an exponential number of calls.

This happens when the main render target is resized through `CompositorWorkspace::_update`. The workspace behaves fine outside of resizing.

`CompositorWorkspace::_update` already calls `CompositorNode::notifyRecreated` (through `CompositorNode::finalTargetResized02`) for all nodes, so I wonder if nodes notifying connected nodes even does anything.

#### Reproducing
Create this workspace and resize it. There should be noticeable lag when resizing.
```
compositor_node Scene
{
	in 0 rt0

	target rt0
	{
		pass clear
		{
			colour_value 0.5 0 0.5 1
		}
	}
}

compositor_node EmptyStart
{
	texture rt0 target_width target_height PFG_RGBA16_FLOAT msaa_auto
	texture rt1 target_width target_height PFG_RGBA16_FLOAT msaa_auto

	out 0 rt0
	out 1 rt1
}

compositor_node Empty
{
	in 0 rt0
	in 1 rt1

	target rt0
	{
		pass clear {}
	}

	target rt1
	{
		pass clear {}
	}

	out 0 rt0
	out 1 rt1
}

workspace ExampleWorkspace
{
	connect_external 0 Scene 0

	alias Empty00 Empty
	alias Empty01 Empty
	alias Empty02 Empty
	alias Empty03 Empty
	alias Empty04 Empty
	alias Empty05 Empty
	alias Empty06 Empty
	alias Empty07 Empty
	alias Empty08 Empty
	alias Empty09 Empty
	alias Empty10 Empty
	alias Empty11 Empty
	alias Empty12 Empty
	alias Empty13 Empty
	alias Empty14 Empty
	alias Empty15 Empty
	alias Empty16 Empty
	alias Empty17 Empty
	alias Empty18 Empty
	alias Empty19 Empty
	alias Empty20 Empty

[...]

PR #555, opened 2025-10-29
Remove recursive calling of CompositorNode::notifyRecreated which cou…

nimble0:
Fixes https://github.com/OGRECave/ogre-next/issues/539

This replaces recursively calling `CompositorNode::notifyRecreated` by following the `CompositorNode::mConnectedNodes` field with notifying a flat list of nodes (`CompositorWorkspace::mNodeSequence`).

Both regular nodes and shadow nodes notify all regular nodes and neither notify shadow nodes. I think this is correct based on this line from the compositor documentation:

> It is possible however, to connect the output from a Shadow Node to a regular Node for further postprocessing (i.e. reflective shadow maps for real time Global Illumination), but Shadow Nodes cannot have input.

`CompositorNode::notifyDestroyed` also recursively notifies other nodes but as it clears the output at the same time it doesn't suffer from the same issue. I've left that behaviour the same although it makes the notify methods a bit inconsistent.

### DIFF
--- a/OgreMain/include/Compositor/OgreCompositorNode.h
+++ b/OgreMain/include/Compositor/OgreCompositorNode.h
@@ -308,9 +308,12 @@ namespace Ogre
         @param finalTarget
             The Final Target (i.e. RenderWindow) from which we'll base our local textures'
             resolution.
+        @param notifyNodes
+            Array of nodes that may be using our textures and need to be notified.
         */
         virtual void finalTargetResized01( const TextureGpu *finalTarget );
-        virtual void finalTargetResized02( const TextureGpu *finalTarget );
+        virtual void finalTargetResized02( const TextureGpu        *finalTarget,
+                                           const CompositorNodeVec &notifyNodes );
 
         /// @copydoc CompositorWorkspace::resetAllNumPassesLeft
         void resetAllNumPassesLeft();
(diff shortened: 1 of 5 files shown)


darksylinc:
Hi! Good job!

Your PR was good, and it made me realize what the problem was (which is why it was [solved differently](https://github.com/OGRECave/ogre-next/commit/b619ad671657ea98445c6b8e228a3e64ff696f3f), and your PR was closed without being merged. Nonetheless your PR was crucial to arrive to this solution).

Your PR had a couple issues:

1. It changed ABI/API which means it cannot be backported to 3.0 (this isn't critical, but if an alternative solution exists, it might be preferred depending on its value / effort).
2. It still had potential O(NxM) (if M = N, then it's N², N is number of nodes, and M is the number of outputs per node) behavior because every node may end up notifying every other node. Chances of this becoming a problem is low though, due to N often being small enough.
3. Your change could've removed mConnectedNodes entirely, since its purpose was to keep track of connections to notify them of changes and avoid problem 2.
4. In OgreCompositorWorkspace.cpp you stop until you find the first `( *itor )->getEnabled()`. This assumes there are no enabled nodes after seeing a disabled one. Unfortunately this might not be the case because advanced users may call `CompositorNode::setEnabled` dynamically, which breaks the assumption (advanced users are in charge of ensuring calling `CompositorNode::setEnabled` directly doesn't produce proper output).

None of these problems were critical or unsolvable; but I believe the alternative solution was superior.

Thanks again! and don't feel your PR was discarded effort. Without it, I probably would've never written the solution I ended up going with.

Cheers

PR #557, opened 2025-11-22
Fix crash when FreeImage library has been compiled without support fo…

nimble0:
…r some formats.

In the case of `FreeImageRe` there might also be gaps even if compiled with full support.


darksylinc:
Thanks!

PR #564, opened 2026-04-01
fix v2-3 for GCC 15

nim65s:
Hi,

upgrading to GCC 15 yield:

```cpp
[50/750] Building CXX object OgreMain/CMakeFiles/OgreNextMain.dir/src/OgreGpuProgramParams.cpp.o
FAILED: [code=1] OgreMain/CMakeFiles/OgreNextMain.dir/src/OgreGpuProgramParams.cpp.o
/nix/store/a245z3cvf9x9sn0xlk6k8j9xhxbhda1z-gcc-wrapper-15.2.0/bin/g++ -DFREEIMAGE_LIB -DOGRE_NONCLIENT_BUILD -DOgreNextMain_EXPORTS -D_MT -D_USRDLL -I/build/source/OgreMain/include -I/build/source/build/include -I/nix/store/d4i1vzn8wsna9rjcc2h55qcsnd8mg815-freetype-2.13.3-dev/include/freetype2 -I/build/source -I/build/source/OgreMain/include/GLX -I/build/source/OgreMain/src/nedmalloc -Wall -Winit-self -Wno-overloaded-virtual -Wcast-qual -Wwrite-strings -Wextra -Wno-unused-parameter -Wshadow -Wno-missing-field-initializers -Wno-long-long -Wno-unused-but-set-parameter  -msse -msse2 -O3 -DNDEBUG -fPIC   -fPIC -pthread -DOGRE_GCC_VISIBILITY -fvisibility=hidden -fvisibility-inlines-hidden -MD -MT OgreMain/CMakeFiles/OgreNextMain.dir/src/OgreGpuProgramParams.cpp.o -MF OgreMain/CMakeFiles/OgreNextMain.dir/src/OgreGpuProgramParams.cpp.o.d -o OgreMain/CMakeFiles/OgreNextMain.dir/src/OgreGpuProgramParams.cpp.o -c /build/source/OgreMain/src/OgreGpuProgramParams.cpp
In file included from /nix/store/mjf8jlq9grydcdvyw6hb063x5c34g5gf-gcc-15.2.0/include/c++/15.2.0/bits/hashtable.h:37,
                 from /nix/store/mjf8jlq9grydcdvyw6hb063x5c34g5gf-gcc-15.2.0/include/c++/15.2.0/bits/unordered_map.h:33,
                 from /nix/store/mjf8jlq9grydcdvyw6hb063x5c34g5gf-gcc-15.2.0/include/c++/15.2.0/unordered_map:43,
[...]

Issue #541, opened 2025-09-12
Traits changes in libc++ 9 break compilations with ogre-2.3.3


j-rivero:
#### System Information
- Ogre Version: 2.3.3
- Operating System / Platform: OSX via Conda
- RenderSystem: -
- GPU: -

#### Detailled description

The gz-rendering compilation seems to be broken on Conda using ogre-2.3.3 for Mac, see https://github.com/conda-forge/gz-rendering-feedstock/pull/42#issuecomment-3221195259.

I think that the problem is a change in the libc++ ,[ concretely in 19 version](https://releases.llvm.org/19.1.0/projects/libcxx/docs/ReleaseNotes.html#deprecations-and-removals);

> The base template for std::char_traits has been removed in LLVM 19. If you are using std::char_traits with types other than char, wchar_t, char8_t, char16_t, char32_t or a custom character type for which you specialized std::char_traits, your code will stop working. The Standard does not mandate that a base template is provided, and such a base template is bound to be incorrect for some types, which could currently cause unexpected behavior while going undetected.

[Line 235 in OgreUTFString](https://github.com/OGRECave/ogre-next/blob/v2.3.3/OgreMain/include/OgreUTFString.h#L235) seems to be triggering the error by a template expansion over [dstring](https://github.com/OGRECave/ogre-next/blob/v2.3.3/OgreMain/include/OgreUTFString.h#L200) that uses a uint32.

Probably the right move would be to transform dstring into `std::u16string` or something that uses `char16_t` instead of `uint16`. But I'm unsure about the implications in maintaining the API/ABI compatibility and if there is a better alternative.

Thanks!

iche033:
thanks for catching this and pointing out that OgreUTFString is only for the v1 Overlay system. We're not really using it so we should be able to remove it in gz-rendering

PR #568, opened 2026-04-09
Fix STLAllocator compatibility with GCC 15 and modern C++ standards

wentasah:
Updated STLAllocator constructors to be compatible with modern C++ standard library containers:

1. Made constructors constexpr: Added constexpr keyword to all constructors to make them compatible with modern C++ standard library containers that require constexpr allocators.

2. Removed explicit from default constructor: The default constructor was marked as explicit, which prevented it from being used in implicit initialization contexts (like {}-initialization). Standard library containers expect allocators to have non-explicit default constructors.

3. Conditionally made destructor constexpr: Added a preprocessor check to only make the destructor constexpr in C++20 and later, since constexpr destructors are not available in earlier C++ standards.

These changes fix compilation errors with GCC 15 where the hashtable implementation tries to use {} initialization on the allocator member, which requires a non-explicit constexpr default constructor.

---

Note: this fixes the same problem as #564, but was entirely AI generated. I'm not sure whether you have some AI policy so feel free to disregard it if it violates your policy.

### DIFF
--- a/OgreMain/include/OgreMemorySTLAllocator.h
+++ b/OgreMain/include/OgreMemorySTLAllocator.h
@@ -100,25 +100,29 @@ namespace Ogre
         };
 
         /// ctor
-        inline explicit STLAllocator()
+        inline constexpr STLAllocator()
         { }
 
         /// dtor
+#if __cplusplus >= 202002L
+        constexpr ~STLAllocator()
+#else
         ~STLAllocator()
+#endif
         { }
 
         /// copy ctor - done component wise
-        inline STLAllocator( STLAllocator const& )
+        inline constexpr STLAllocator( STLAllocator const& )
         { }
 
         /// cast
         template <typename U>
-        inline STLAllocator( STLAllocator<U, AllocPolicy> const& )
+        inline constexpr STLAllocator( STLAllocator<U, AllocPolicy> const& )
         { }
 
         /// cast
         template <typename U, typename P>
-        inline STLAllocator( STLAllocator<U, P> const& )
+        inline constexpr STLAllocator( STLAllocator<U, P> const& )
         { }
 
         /// memory allocation (elements, used by STL)


darksylinc:
What's the take from Gazebo team?

They don't like changes that break the ABI and this almost certainly an ABI-breaking change (I didn't run the abi checker to confirm).

There was a similar discussion with #541 where Apple broke support. Fortunately we avoided it entirely because the affected component wasn't actually in use.

Given that at this point the v2-3 branch is basically "whatever gazebo is doing", it may be solvable by simply doing `#ifdef GCC >= 15` (not actual macro name), thus Gazebo ABI compiled with GCC 15 is compatible with Ogre 2.3 compiled with GCC 15; while Gazebo GCC 14 remains compatible with Ogre 2.3 compiled with GCC 15.

wentasah:
@scpeters Can you answer the above?

We're using this [patch in nix-ros-overlay](https://github.com/lopsided98/nix-ros-overlay/blob/909dbfd8cbe571a6cb63ac7a526e71c8d3f85a26/distros/distro-overlay.nix#L242) and it works fine, but ABI compatibility is not that important in Nix ecosystem as elsewhere, and I have no clue how and where is this used.

scpeters:
> @scpeters Can you answer the above?
> 
> We're using this [patch in nix-ros-overlay](https://github.com/lopsided98/nix-ros-overlay/blob/909dbfd8cbe571a6cb63ac7a526e71c8d3f85a26/distros/distro-overlay.nix#L242) and it works fine, but ABI compatibility is not that important in Nix ecosystem as elsewhere, and I have no clue how and where is this used.

I will discuss with the gazebo team

cc @azeey @iche033 @j-rivero

j-rivero:
> What's the take from Gazebo team?

Thanks for checking. Given our model of loading plugins for gz-rendering, we can probably deal with an ABI breakage without too much problem for the end users that load the gz-rendering-ogre2 plugin directly from our binaries.

Said that, I known that we declared the Semantic Versioning starting with 3.x but might be a good practice try to keep it stable if we can. For this case, we can potentially add a compiler flag (default to OFF) that brings the patches in this PR (something like -DUSE_MODERN_CXX) so the people that needs the patch enable it in their building metadata while the software maintains the ABI stability for all the 2.3 potential consumers out there.

wentasah:
> we can potentially add a compiler flag (default to OFF) that brings the patches in this PR (something like -DUSE_MODERN_CXX) so the people that needs the patch enable it in their building metadata while the software maintains the ABI stability for all the 2.3 potential consumers out there

I'm fine with adding `-DUSE_MODERN_CXX`, but I was curious about the ABI incompatibilities and run the change through [abi-compliance-checker](https://github.com/lvc/abi-compliance-checker) and it [reported 100% compatibility for libOgreMain.so](https://github.com/user-attachments/files/26813076/compat_report.html). The result for other libraries was the same. If the tool can be trusted, changes seem not to be necessary. But the tool was last updated 5 years ago, so I can image it might have missed some incompatibilities related to modern C++. Does anybody have experience with reliability of abi-compliance-checker reports?

j-rivero:
> . Does anybody have experience with reliability of abi-compliance-checker reports?

We have been using it for the last decade in Gazebo and it we did not have any serious problem that it was not detected. So it is old but still works I would say.

Looking now more into the changes i think that they affect to the source semantics but not to the binary layer generated to my eyes.

darksylinc:
Thanks! Give. That 2.3 branch maintenance has essentially become "whatever gazebo" needs, if this breaks something you can submit another PR

PR #580, opened 2026-06-22
MeshLodGenerator: native v2 Mesh support

Laxx18:
**Quick summary of what it does and what I found along the way:**

Two new classes (LodInputProviderMeshV2, LodOutputProviderMeshV2) slot into the existing LodInputProvider/LodOutputProvider strategy pattern, so MeshLodGenerator can read/write directly against a v2 Mesh's SubMesh Vaos — no v1 mesh involved anywhere.

Turns out the runtime side (camera-distance LOD switching for Item) already works end-to-end for v2 — I'd assumed otherwise partway through and want to flag that I was wrong about that, in case it's useful context. The actual gap was purely on the generation side.

MeshLodUpgrader now generates LOD for v2-native source meshes too, not just v1.
Added a sample (Sample_MeshLodV2) comparing a procedurally-built v2 mesh against Sinbad (v1-imported-then-LOD-generated-natively) side by side.

Three things I'd specifically like your read on before going further:

1. LodConfig's API shape — I added a parallel meshV2 field rather than changing the existing mesh field's type, to avoid breaking the v1 API. Open to a different shape if you'd prefer.

2. Manual (mesh-swap) LOD levels and independent (non-shared) shadow-mapping Vaos are explicitly unsupported for the v2 path in this first pass — both fail loudly rather than silently doing the wrong thing. Worth scoping into a follow-up, or should this PR cover them?

3. Mesh::removeLodLevels() was previously a no-op stub (commented-out body) — I implemented it for real since MeshTool's "drop LOD" option needs it to actually work. Wanted to flag that explicitly since it's existing dead code I'm reviving, not something I added net-new.

Here the result:

Result

Best Regards
Lax


Components/Hlms/Pbs/src/OgreHlmsPbs.cpp
```diff
@@ -3128,6 +3128,8 @@ namespace Ogre
 
         if( OGRE_EXTRACT_HLMS_TYPE_FROM_CACHE_HASH( lastCacheHash ) != mType )
         {
+            if( mCurrentPassBuffer == 0 )
```
darksylinc: This is unrelated.

darksylinc (changes requested):
Everything else looks alright. Good job!

PR #583, opened 2026-07-16
Fixing the D3D11RenderSystem::createBlendState and Hlms::applyStrongBlendblockRules

bishopnator:
There seems to be a problen D3D11 that the blendblock's blend operations were not properly setup according to mSeparateBlend and in HLMS implementation the flag was not udpated after applying the strong blendblock rules.


darksylinc:
Oh my. It was completely broken.

Thanks!

